PR #1, opened 2019-05-13
Creates initial project boilerplate

robmarch2:


CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+an individual is representing the project or its community in public spaces.
+Examples of representing a project or community include using an official
+project e-mail address, posting via an official social media account, or acting
+as an appointed representative at an online or offline event. Representation of
+a project may be further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at [TODO INSERT EMAIL ADDRESS]. All
```
robmarch2: This needs a real email address

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
+# Contributor Covenant Code of Conduct
```
garyttierney: cc @danielgrant what's the intention here?

build.gradle.kts
```diff
@@ -0,0 +1,55 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+
+    dependencies {
+        classpath("org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:2.7")
```
garyttierney: Why are these plugin dependencies duplicated? They already exist in the `plugins` block.

Dockerfile.build
```diff
@@ -1,7 +1,7 @@
 FROM gradle:5.4.1-jdk11
 LABEL maintainer="Rob March <rob.march@digirati.com>"
 
-RUN apt-get clean && apt-get update && apt-get -y upgrade
+RUN apt-get clean && apt-get update && apt-get -y upgrade \
```
garyttierney: `Dockerfile` linting should be added to our pre-commit config.

build.gradle.kts
robmarch2: Do they not need to be imported as buildscript dependencies before applying them in the `plugins` block?

garyttierney: Nope, though I see where the confusion might come from. In Gradle 4, you had to add plugins as dependencies to the buildscript classpath:

```
buildscript {
  dependencies {
    classpath("xyz")
  }
}

apply plugin: "xyz"
```

Since Gradle 5, using the `plugins { ... }` syntax handles both for you:
```
plugins {
  id("xyz")
}
```

Dockerfile.build
```diff
@@ -0,0 +1,9 @@
+FROM gradle:5.4.1-jdk11
+LABEL maintainer="Rob March <rob.march@digirati.com>"
+
+USER root
```
garyttierney: Why are we doing this?

Dockerfile.build
garyttierney: Hadolint is a good tool, here's a pre-commit config for it: https://github.com/stratasan/hadolint-pre-commit

Dockerfile.build
robmarch2: The `apt-get` commands were failing due to not having the appropriate permissions (e.g. https://jenkins.build.digirati.com/job/digirati-taxonomy-manager/job/PR-1/3/console ). The solutions to it that I saw online all suggested running the command as root but this seemed like a cleaner solution than putting `sudo`s for each of the `apt-get` commands

CONTRIBUTING.md
robmarch2: This is to comply with the open source RFC - https://github.com/digirati-co-uk/rfc/tree/master/rfc/rfc2---open-source-standards#contributing

Dockerfile.build
garyttierney: We should switch back to the original user before we actually run the commands (e.g., switch to root, install packages, switch back, run build). It's relatively trivial to escape a container as root.

CONTRIBUTING.md
garyttierney: Yep, this is more one for @danielgrant about why this is in there in the first place.

Dockerfile.build
garyttierney: Er, when I say _the commands_, I mean the build commands. So our `./gradlew build` or whatever.

CONTRIBUTING.md
danielgrant: It's part of an effort to make projects more appealing and inviting to contributors, and also forms part of GitHub's Open Source Guide (https://opensource.guide/). You can view project alignment to this guidance under "Insights" -> "Community" (i.e. https://github.com/dlcs/digirati-taxonomy-manager/community).

That being said, I'm not averse to removing it if someone can make a case for why it shouldn't be in there.

CONTRIBUTING.md
danielgrant: opensource@digirati.com

Dockerfile.build
danielgrant: I prefer using `sudo` rather than switching to `root`, even it's only during the image build.

Dockerfile.build
```diff
@@ -0,0 +1,9 @@
+FROM gradle:5.4.1-jdk11
+LABEL maintainer="Rob March <rob.march@digirati.com>"
```
danielgrant: Let's switch this to `Digirati <opensource@digirati.com>`.

Dockerfile.build
garyttierney: `sudo` might not be available in the image. In fact, it shouldn't be.

danielgrant: We can install it for the purposes of building the image, and if we use a multi-stage build we don't have to carry that dependency over to the final image.

garyttierney: I'm not understanding why we would install an external package from inside the build to switch UIDs when Docker supports doing it for us.

Dockerfile.build
```diff
@@ -0,0 +1,11 @@
+FROM gradle:5.4.1-jdk11
```
garyttierney: We don't actually need a `gradle` image since we're running the build via the wrapper.

LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2019 Digirati Ltd
```
danielgrant: Just `Digirati` :smile:

Jenkinsfile
```diff
@@ -0,0 +1,72 @@
+pipeline {
+    agent {
+        dockerfile {
+            filename 'Dockerfile.build'
+        }
+    }
+
+    stages {
```
garyttierney: I haven't quite thought about how to approach this yet, however, we can share parts of Jenkinsfiles between projects (like so: https://github.com/digirati-co-uk/dlcs-environment-pipeline/tree/master/vars). I'd like to do that so we can reduce the duplication and maintenance burden of having [relatively] large Jenkinsfiles.

E.g., we could have something like:
```
genericCodeAnalysis(['javac', 'spotbugs', 'checkstyle'])
```
And we can start using defaults where appropriate.

robmarch2: I don't have permissions to view that, but if it's anything like this I'd be down for that:

https://jenkins.io/blog/2017/10/02/pipeline-templates-with-shared-libraries/

garyttierney: Ah, yep, that's it. You should be able to access that repo now. They work quite well since you can write automated tests around the pipelines themselves if the steps get large.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+-   repo: https://github.com/kintoandar/pre-commit.git
+    sha: v2.1.0
+    hooks:
```
garyttierney: Doesn't need to be here.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: hadolint
```
garyttierney: You dropped this in, but never referenced it in the pre-commit-config file.

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 
--   repo: .
-    rev: 4b5b03cd59316dc69c1e178fccce0be6101b644c
+-   repo: git://github.com/stratasan/hadolint-pre-commit
```
garyttierney: Looks like you're pointing to a remote because you couldn't get the local hooks working. See: https://pre-commit.com/#repository-local-hooks

robmarch2: Ah, fantastic. I'll switch back once I've managed to get hadolint to actually run. The problem at the moment is that it needs to run docker from within the docker build agent

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,14 @@
+FROM openjdk:11
+LABEL maintainer="Digirati <opensource@digirati.com>"
+
+USER root
```
danielgrant: I don't think we need to switch to `root` here - the user defaults to `root` for the `openjdk:11` image.

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,14 @@
...
+
+RUN apt-get clean && apt-get update && apt-get -y upgrade \
```
danielgrant: `apt-get clean` should come last, followed by `rm -rf /var/lib/apt/lists/*`.

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,14 @@
...
+    && apt-get install -y python python-pip software-properties-common lsb-release apt-transport-https \
```
danielgrant: Let's target specific versions here. We can find the versions currently available using `apt-cache madison <package-name>`.

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,14 @@
...
+    && curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
+    && add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable" \
+    && apt-get update && apt-get -y install docker-ce
+
+RUN pip install pre-commit
```
danielgrant: We should target the specific version of `pre-commit` we want to install, this can be found using `pip search <package-name>`.

Also, I suggest rolling this `pip install` into the above `RUN` to keep the build image size down.

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,14 @@
...
+
+USER gradle
```
danielgrant: We're switching to the `gradle` user before it's actually been created. This doesn't cause an error during `docker build`, but any attempt to actually use this user will result in `unable to find user gradle: no matching entries in passwd file`.

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,14 @@
+FROM openjdk:11
```
stephenwf: Any reason for the Dockerfile not being in the root of the project?

robmarch2: This is mostly because we're likely to end up with several so it seemed like a good idea to not leave them in the root. Also, it makes the dockerfile linting step in the Jenkinsfile a bit simpler - we can just tell it to lint everything in `/dockerfiles` rather than hard-coding the names of each dockerfile into the pipeline as (I think) we'd have to do if we just left them all in the root

robmarch2:
> Remove the Docker installation and its dependencies (lsb-release, apt-transport-https, software-properties-common).

We can't do this. I tried in a previous commit and it stopped hadolint from running (see https://jenkins.build.digirati.com/job/digirati-taxonomy-manager/view/change-requests/job/PR-1/28/console )

> Is Ubuntu not shipping Python3?

It seems to be a different dependency to just `python` - the version chosen here was the one that was being installed, but I'm happy to try and move it over to python3

garyttierney:
> We can't do this. I tried in a previous commit and it stopped hadolint from running (see https://jenkins.build.digirati.com/job/digirati-taxonomy-manager/view/change-requests/job/PR-1/28/console)

Hm. This is odd. The `dockefiler linting` step is running inside the container created by `Dockerfile.build`, but it shouldn't be nesting containers when you switch agents. Let's leave it for now and refactor this pipeline into something tidier later.

PR #3, opened 2019-05-16
Create wrapper around Aho-Corasick impl and term-concept lookup

robmarch2:


src/main/java/com/digirati/taxonomy/manager/lookup/model/Match.java
```diff
@@ -0,0 +1,48 @@
...
+ * Represents a pairing between the details of where a term was found in a piece of input text, and
+ * what that term could mean.
+ */
+public class Match {
+
+    private final Emit emit;
+
+    private final List<Concept> matchingConcepts;
+
+    public Match(Emit emit, List<Concept> matchingConcepts) {
```
garyttierney: You use `java.util.Collection` in most other places, but have a `java.util.List` here. Any particular reason why?

src/test/java/com/digirati/taxonomy/manager/lookup/AhoCorasickAutomatonTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test
+    void searchShouldNotMatchTermsSeparatedByPunctuation() {
+        // Given
+        AhoCorasickAutomaton underTest = new AhoCorasickAutomaton("aho", "corasick");
+
+        // When
+        Collection<Emit> matches = underTest.search("aho-corasick");
+
+        // Then
+        MatcherAssert.assertThat(matches, Matchers.iterableWithSize(0));
```
garyttierney: Again, less clear than just `assertTrue(matches.isEmpty())`

build.gradle.kts
```diff
@@ -17,6 +17,13 @@ java {
 
 dependencies {
     spotbugsPlugins("com.h3xstream.findsecbugs:findsecbugs-plugin:1.7.1")
+
+    compile("org.ahocorasick", "ahocorasick", "0.4.0")
```
garyttierney: Versions should be in `gradle.properties`.

src/main/java/com/digirati/taxonomy/manager/lookup/Lookup.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param terms the terms to pass into the reconstructed automaton
+     */
+    public void rebuildAutomaton(String... terms) {
+        ahoCorasickAutomaton = new AhoCorasickAutomaton(terms);
+    }
+
+    /**
+     * Replaces the current {@link ConceptLookupTable} with an empty new one. When changes need to
+     * be made to the structure of terms and concepts, updating the existing lookup table is likely
+     * to be error prone - instead we simply rebuild it from scratch
```
garyttierney: > updating the existing lookup table is to be error prone.

In what way?

src/test/java/com/digirati/taxonomy/manager/lookup/LookupTest.java
```diff
@@ -0,0 +1,117 @@
...
+        assertMatches(
+                matches,
+                match("row", 2, 4),
+                match("sake", 24, 27),
+                match("row", 62, 64),
+                match("sake", 69, 72));
+    }
+
+    private void assertMatches(Collection<Match> actual, Match... expected) {
+        MatcherAssert.assertThat(actual, Matchers.contains(expected));
```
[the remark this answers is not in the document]
robmarch2: `Matchers.contains` is a badly worded method name -it's actually asserting that the lists contain these and only these, in the order they're provided. The one for checking a subset is `Matchers.hasItems`

src/test/java/com/digirati/taxonomy/manager/lookup/LookupTest.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private Match match(String term, int startIndex, int endIndex) {
+        return new Match(new Emit(startIndex, endIndex, term), null);
+    }
+
+    private Match match(String term, int startIndex, int endIndex, Concept... concepts) {
+        return new Match(new Emit(startIndex, endIndex, term), Lists.newArrayList(concepts));
```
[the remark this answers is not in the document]
robmarch2: The reason for this is the same as the reason for the null above - when no matching concepts are found in the concept lookup table, the Match has a null for the list of concepts but this puts an empty list which, with the production code as it currently is, will cause this `Match` to not be equal to the actual one. I had avoided changing the production code just to cater for the tests, but because this is new code I'm not too precious about that point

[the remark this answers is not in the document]
garyttierney: So the tests only pass here because production code allows you to use the API incorrectly. The `Match` constructor should be changed to reject `null` as a value.

src/main/java/com/digirati/taxonomy/manager/lookup/model/Match.java
robmarch2: The reason here is just that I happened to use `List` in the `ConceptLookupTable`. I don't see any reason why that has to be a `List` there so I'm happy to change it

build.gradle.kts
robmarch2: I understand why we want to do that where we have multiple subprojects and want to share the same version between all projects, but in cases like this where we have a single gradle project what benefit do we get from this?

src/test/java/com/digirati/taxonomy/manager/lookup/LookupTest.java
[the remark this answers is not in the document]
garyttierney: Hm, that's not clear at all.

build.gradle.kts
garyttierney: We _will_ have multiple subprojects under this repository. Building several restful APIs is part of this product.

Besides that, it forces you to think about the current version you're using rather than copying the `Gradle` section from mvnrepository.com.

src/main/java/com/digirati/taxonomy/manager/lookup/Lookup.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Replaces the current {@link AhoCorasickAutomaton} with a new one. When changes are made to
+     * what terms should be listed in it, it isn't possible to update the existing automaton with
+     * those changes without a lot of very error-prone work - instead we simply rebuild it from
+     * scratch
+     *
+     * @param terms the terms to pass into the reconstructed automaton
+     */
+    public void rebuildAutomaton(String... terms) {
+        ahoCorasickAutomaton = new AhoCorasickAutomaton(terms);
```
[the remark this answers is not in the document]
robmarch2: I don't think there will be, but I'm unsure of what form the data from which we build the lookup table will take - but it'll definitely need more than a collection of Strings. For now I figured I'd leave it as something for the calling code to do because if we were to add the call to clear the existing lookup table into this method we'd end up in the exact same state as merging these is trying to solve - half of the state would be correct and up to date and the other half wouldn't be. Once we've got an understanding of how that works, we should probably merge the two into one, but not until then.

build.gradle.kts
garyttierney: Actually, I'd be inclined to drop this for something like `dependabot`

garyttierney: (cc @danielgrant)

src/main/java/com/digirati/taxonomy/manager/lookup/model/Match.java
garyttierney: Unless you do need an indexed and ordered collection of items, returning `Collection` is more appropriate.

build.gradle.kts
stephenwf: Could use [Renovate](https://github.com/renovatebot/renovate), used quite well on other Frontend OSS. Can be configured to auto-merge patch versions so long as the tests pass.

src/main/java/com/digirati/taxonomy/manager/lookup/Lookup.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Constructor creates all data structures necessary for extracting {@link
+     * com.digirati.taxonomy.manager.lookup.model.Concept}s from a given piece of text. At present
+     * this consists of an {@link AhoCorasickAutomaton} which is used to locate terms in text, and a
+     * {@link ConceptLookupTable} which is used to convert term matches into {@link
+     * com.digirati.taxonomy.manager.lookup.model.Concept} matches.
+     *
+     * @param terms all possible terms to look for
+     */
+    public Lookup(String... terms) {
```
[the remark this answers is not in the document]
garyttierney: Also, think about inversion of control. We should be passing down an `AhoCorasickAutomaton` and a `ConceptLookupTable`, not the required information for them to be created.

src/main/java/com/digirati/taxonomy/manager/lookup/Lookup.java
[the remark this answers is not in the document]
garyttierney: I think we can just remove these "rebuild" APIs and make those fields immutable.

src/main/java/com/digirati/taxonomy/manager/lookup/ConceptLookupHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.collect.Multimap;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Encapsulates all operations related to determining which concepts exist within a given piece of
+ * text.
+ */
+public class ConceptLookupHandler {
```
garyttierney: Hm, it does do concept lookups for a given text, but the "Handler" suffix gives the impression that it's a sort of event listener, or reacts to a request.

What do you think about `TextClassifier`, or `TextCategorizer` (and associated `ConceptLookupHandler#search` -> `TextClassifier|Categorier#classify|categorize`)?

cc @p-kaczynski @danielgrant for bikeshedding some names.

robmarch2: I couldn't think of a good name in the moment (hence this not great one) but now that I've pushed it up I've just had a thought - how about `ConceptExtractor#extract`?

garyttierney: Yep, that sounds good to me.

PR #6, opened 2019-05-29
Creates text normalisation mechanism

robmarch2:


src/main/java/com/digirati/taxonomy/manager/normalisation/TextNormaliser.java
```diff
@@ -0,0 +1,74 @@
...
+     * we're using, but language-specific tokenisation should still work
+     *
+     * @param stopwords a collection of common boilerplate words to remove from the text (e.g.
+     *     "the", "a", etc.)
+     * @param languageKey the ISO 639-1 language code for the text to be normalised
+     * @param languageName the name of the language of the text to be normalised
+     */
+    public TextNormaliser(Collection<String> stopwords, String languageKey, String languageName) {
+        this.stopwords = stopwords;
+        this.pipeline = createPipeline(languageKey, languageName);
```
garyttierney: This call-chain (`createPipeline`) does an expensive set up of the annotation pipeline. This needs to happen somewhere where the computational cost is obvious. The expectation of the `new` operator is that it only allocates for the object you're creating.

build.gradle.kts
```diff
@@ -26,6 +27,9 @@ dependencies {
 
     compile("org.ahocorasick", "ahocorasick", ahoCorasickVersion)
     compile("com.google.guava", "guava", guavaVersion)
+    compile("edu.stanford.nlp", "stanford-corenlp", stanfordNlpVersion)
+    compile("edu.stanford.nlp:stanford-corenlp:$stanfordNlpVersion:models")
+    compile("edu.stanford.nlp:stanford-corenlp:$stanfordNlpVersion:models-french")
```
garyttierney: Why are we depending on a french model?

src/main/java/com/digirati/taxonomy/manager/normalisation/TextNormaliser.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Collection;
+import java.util.Properties;
+
+/**
+ * Provides a mechanism for normalising any text inputs, including the terms to search for and the
+ * text in which to search.
+ */
+public class TextNormaliser {
+
+    private final Collection<String> stopwords;
```
garyttierney: This should be a `Set`. `Set<T>#contains(T)` is O(1).

src/test/java/com/digirati/taxonomy/manager/normalisation/TextNormaliserTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class TextNormaliserTest {
+
+    @Test
+    void normaliseShouldLemmatizeText() {
```
garyttierney: We shouldn't be testing specific lemmatization results, that can be changed on a whim by the underlying library (and we don't particularlycare if it does).

build.gradle.kts
robmarch2: Oops. This was a hangover from when I was trying to get French lemmatisation working. Will remove

src/test/java/com/digirati/taxonomy/manager/normalisation/TextNormaliserTest.java
robmarch2: These were about trying to document the behaviour of the library we're relying on, but I take your point that we don't really care about the specifics of it. I'll drop the tests about contractions and US/UK english, but since we're relying on the fact that some sort of lemmatisation occurs I think we should leave the happy-path test in place.

stephenwf: If the libraries we are using change the lemmatization results, surely that's a breaking change, I'd vote to keep these, even if they act as a way to flag up changes in the underlying library when we come to upgrade.

As you say @robmarch2 documenting which parts of the library we are using would help someone if they came to swap it out too 👍

robmarch2: @stephenwf I think the point @garyttierney is making is that as long as both the taxonomy terms and the input text get lemmatised according to the same scheme, the details of how that lemmatisation occurs (e.g. "I'm" -> "I be", "colors colours" -> "color colour", etc) shouldn't break any of our functionality. 

As far as I can tell, we'll only start getting issues if lemmatisation completely stops working (e.g. if we have "colour" in the taxonomy and "colours" in the input text, if lemmatisation stops working those terms won't match any more, whereas if the library were to change to lemmatise both to the American spelling they still will). So my view is that we need a test that fails in that case, but failures in the ones documenting the inner workings of the library won't help in reasoning about functional problems so we can drop them without losing anything

garyttierney: This is a case where interacting with a mock is fine, and likely preferable. We don't care _what_ the result was, just that we got the result and we used it.

robmarch2: I agree that using a mocked out pipeline would be fine for most of these tests that don't care about the output (e.g. `normaliseShouldTrimWhitespace`), but I want there to be tests that explicitly prove that the lemmatisation step of the pipeline occurs for English but not French since that's a fundamental part of what this class does. I've had a look into the `StanfordCoreNLP` class and I don't think we'll be able to prove that with mocks so I think we should definitely keep the English and French happy path tests as they currently are - I'm happy to change the other 4 over though if you feel strongly about it.

src/main/java/com/digirati/taxonomy/manager/normalisation/TextNormaliser.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Provides a mechanism for normalising any text inputs, including the terms to search for and the
+ * text in which to search.
+ */
+public class TextNormaliser {
+
+    private final Set<String> stopwords;
+
+    private StanfordCoreNLP pipeline;
```
garyttierney: Should still be `final`.

src/main/java/com/digirati/taxonomy/manager/normalisation/TextNormaliser.java
```diff
@@ -0,0 +1,80 @@
...
+     * Initialises a {@link TextNormaliser} for normalising any input language. Currently, full
+     * lemmatisation is only supported for English due to a restriction of the NLP library that
+     * we're using, but language-specific tokenisation should still work
+     *
+     * @param stopwords a collection of common boilerplate words to remove from the text (e.g.
+     *     "the", "a", etc.)
+     * @param languageKey the ISO 639-1 language code for the text to be normalised
+     * @param languageName the name of the language of the text to be normalised
+     */
+    public static TextNormaliser initialiseNormaliser(
```
garyttierney: This isn't really much different from running this in the constructor. If we return a `Future<TextNormaliser>` and run it on a single threaded `Executor` it'll be much more clear that `new StanfordCoreNLP()` is an expensive call.

garyttierney: Really, we just want a way to say "this is a long, blocking operation" -- hiding the initialization behind concurrent APIs to do that is fine, I think.

garyttierney:
This needs rebased on DTM-48.

src/main/java/com/digirati/taxonomy/manager/normalisation/TextNormaliser.java
robmarch2: This can't be final as things currently stand - it gets set in `initialiseNormaliser`. I'll make sure to move it back into the constructor and mark it as final when I action the point about that method though.

robmarch2:
> This needs rebased on DTM-48.

This change came first so 48 is based on an earlier cut of this. I need to rebase 48 onto this branch when all the feedback on here has been actioned.

PR #7, opened 2019-05-29
API specification

garyttierney:


docs/api/samples/concept-scheme.json
```diff
@@ -0,0 +1,14 @@
...
+        "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+        "skos": "http://www.w3.org/2004/02/skos/core#",
+        "xsd": "http://www.w3.org/2001/XMLSchema#"
+    },
+    "@id": "http://example.com/taxonomy",
+    "dcterms:title": "Taxonomy Title",
+    "dcterms:description": "This is the description of my taxonomy!",
+    "dcterms:created": "2019-08-01",
+    "dcterms:updated": "2019-08-01"
```
stephenwf: Will a concept scheme contain any of the top level concepts inside the `GET` request?

docs/api/management.oas3.yaml
```diff
@@ -0,0 +1,140 @@
...
+            example:
+              $ref: './samples/concept.json'       
+      responses:
+         '204':
+            description: Update was successful, and the skos:Concept is unchanged from the request body.              
+
+  /0.1/concept-scheme/{scheme}/concept:
+    parameters:
+      - $ref: '#/components/parameters/scheme'
+    post:
```
stephenwf: It would be useful to have a `GET` on this, to see all of the concepts inside of a scheme, direct relationships from `skos:inScheme` terms on the scheme.

docs/api/management.oas3.yaml
```diff
@@ -0,0 +1,140 @@
...
+          content:
+            'application/ld+json;profile="http://www.w3.org/2004/02/skos/core.jsonld"':
+              example:
+                $ref: './samples/concept-in-scheme.json'
+
+  /0.1/concept-scheme/{scheme}:
+    parameters:
+      - $ref: '#/components/parameters/scheme'
+    get:
+      summary: Get a ConceptScheme
```
stephenwf: In addition to `GET` on the scheme, we need a way to update and delete a single scheme

docs/api/management.oas3.yaml
```diff
@@ -0,0 +1,140 @@
...
+          $ref: '#/components/schemas/error'
+        '200':
+          description: A ConceptScheme resource.
+          content:
+            'application/ld+json;profile="http://www.w3.org/2004/02/skos/core.jsonld"':
+              example: 
+                $ref: samples/concept-scheme.json
+
+  /0.1/concept-scheme/:
+    post:
```
stephenwf: The API may also need to expose a `GET` on this endpoint to list all of the available concept scheme, to be used as an entry point into the API

docs/api/management.oas3.yaml
```diff
@@ -0,0 +1,140 @@
...
+            'application/ld+json;profile="http://www.w3.org/2004/02/skos/core.jsonld"':
+              example: 
+   
+components: 
+
+  responses:
+    notFound:
+      description: The specified resource was not found
+      content:
+        application/json:
```
stephenwf: We could use `application/problem+json`

docs/api/samples/concept-scheme.json
```diff
@@ -0,0 +1,14 @@
+{
+    "@context": {
+        "dc": "http://purl.org/dc/elements/1.1/",
+        "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+        "skos": "http://www.w3.org/2004/02/skos/core#",
+        "xsd": "http://www.w3.org/2001/XMLSchema#"
+    },
+    "@id": "http://example.com/taxonomy",
```
stephenwf: Scheme doesn't conform to variable represented in `management.oas3.yaml#/parameters/concept` may need another identifier for this, or for the domain?

docs/api/samples/concept-scheme.json
garyttierney: Just in the same way that a `Concept` contains its `narrower`/`broader` properties and we give you the minimal required presentation information:

```
{
    "@type": "skos:ConceptScheme",
    "skos:hasTopConcept": [
         { "@id": "xxx", "skos:prefLabel": { ... } } 
     ]
}
```

docs/api/management.oas3.yaml
```diff
@@ -0,0 +1,140 @@
...
+      summary: Add a skos:Concept to an existing skos:ConceptScheme.
+      operationId: addConceptToScheme
+      requestBody:
+        description: A skos:Concept resource.
+        content:
+          'application/ld+json;profile="http://www.w3.org/2004/02/skos/core.jsonld"':
+            example:
+              $ref: './samples/concept.json'
+      responses:
+        '201':
```
stephenwf: For validation errors we may need a `406` or generic `400` (applied to other `POST` / `PUT` requests)

docs/api/samples/concept-in-scheme.json
```diff
@@ -0,0 +1,19 @@
...
+      "dc": "http://purl.org/dc/elements/1.1/",
+      "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+      "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+      "skos": "http://www.w3.org/2004/02/skos/core#",
+      "xsd": "http://www.w3.org/2001/XMLSchema#"
+    },
+    "@id": "urn:wbs:47",
+    "@type": "skos:Concept",
+    "skos:broader": [{
+      "@id": "urn:wbs:42"
```
stephenwf: Will the linked schemes contain a `skos:prefLabel` here too?

docs/api/samples/concept-in-scheme.json
```diff
@@ -0,0 +1,19 @@
...
+    }],
+    "skos:inScheme": [{
+      "@id": "myscheme"
+    }],
+    "skos:prefLabel": "Trajectory analysis"
```
stephenwf: Are the `skos:prefLabel` going to come back normalised as a language map (with the default language)?

docs/api/samples/concept-in-scheme.json
garyttierney: I hadn't planned on it, but if you want to have a sort of "Appears in these schemes: <X>, <Y>, <Z>" I'll add it in.

docs/api/samples/concept-in-scheme.json
garyttierney: Yes, normalised as a language map. You'll need to be able to determine the current language somehow (JWT?). I should update this example to reflect that.

docs/api/management.oas3.yaml
garyttierney: That's modelled under `GET /0.1/concept-scheme/my-concept-id`, thoughts?

stephenwf: Yeh that's fine, we only need one 👍

docs/api/samples/concept-in-scheme.json
stephenwf: That property is for `skos:broader` but I think in general wherever we've got a link off with an `@id` to a concept or concept-scheme we should have at least the `skos:prefLabel`.

docs/api/samples/concept-in-scheme.json
stephenwf: MVP will just be configured at build time in the FE, but yeh, will be up to the FE to display the right one and fallback as needed.

docs/api/samples/concept-in-scheme.json
garyttierney: I'll add a note to the top of the API spec that _ALL_ embedded links _MUST_ provide all available lexical labels (prefLabel, altLabel, hiddenLabel).

docs/api/schemas/concept-scheme/embedded.json
```diff
@@ -0,0 +1,34 @@
...
+            "examples": [
+                "http://taxman.com/0.1/concept-scheme/1"
+            ],
+            "pattern": "^(.*)$"
+        },
+        "@type": {
+            "$id": "#/skos:ConceptScheme/embedded/@type",
+            "const": "skos:ConceptScheme"
+        },
+        "skos:prefLabel": {
```
robmarch2: Should this not be `dcterms:title`?

docs/api/schemas/concept/embedded.json
```diff
@@ -0,0 +1,34 @@
...
+            "examples": [
+                "http://taxman.com/0.1/concept-scheme/1"
+            ],
+            "pattern": "^(.*)$"
+        },
+        "@type": {
+            "$id": "#/skos:Concept/embedded/@type",
+            "const": "skos:Concept"
+        },
+        "skos:prefLabel": {
```
robmarch2: Looks like a lot of the fields are missing here, as well as in the other concept schemas

docs/api/schemas/concept-scheme/request.json
```diff
@@ -0,0 +1,31 @@
+{
+    "$id": "#/skos:ConceptScheme/request",
+    "type": "object",
+    "title": "skos:ConceptScheme respoonse",
```
robmarch2: Typo here

docs/api/schemas/concept/embedded.json
garyttierney: This is the embedded schema (i.e. used to present a list of concepts), which is why it only has preferred label, and ID.

docs/api/management.oas3.yaml
```diff
@@ -0,0 +1,342 @@
...
+          description: JSON-LD representation of the new Project resource.
+          content:
+            'application/ld+json;profile="http://www.w3.org/2004/02/skos/core.jsonld"':
+              example:
+                $ref: samples/project/new-project--response.json
+              schema:
+                $ref: schemas/project/response.json
+
+  /0.1/concept-scheme/{scheme}/concept/{concept}:
```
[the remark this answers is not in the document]
garyttierney: Both, really. This is something I've touched on briefly with @stephenwf. The frontend currently uses the latter, but the former is the canonical URI for a concept.

Originally this was written with the intention of a concept being associated exclusively to exactly 1 concept scheme, but that's no longer the case. Since a concept can exist across multiple schemes, a way to address that concept unambiguously is needed which is why we have `/v0.1/concept/{concept}`.

For the sake of ease of use, while integrating with the frontend we'll add `/0.1/concept-scheme/{scheme}/concept/{concept}`, which will be functionally equivalent with the addition of attaching the new concept to the concept scheme as a top concept.

(cc @stephenwf).

garyttierney:
> What's the plan with the delete endpoints? Put them into this before merging or add them in a separate PR?

I think we should omit them from this PR, and create a new PR which contains the spec along with the implementation (which would be written after the spec is reviewed, or maybe before if it's trivial enough to not worry about major spec changes).

docs/api/schemas/concept/embedded.json
robmarch2: Fair enough on this one, but the same thing applies to `request.json` and `response.json` - is that deliberate?

docs/api/schemas/concept-scheme/embedded.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$id": "#/skos:ConceptScheme/embedded",
+    "type": "object",
+    "title": "embedded skos:ConceptScheme",
+    "required": [
+        "@id",
+        "@type",
+        "skos:prefLabel"
```
robmarch2: This should be `"dcterms:title"`

docs/api/schemas/concept/embedded.json
garyttierney: Nope, I'm updating those with all the properties.

PR #9, opened 2019-05-30
Map concept matches from normalised form back to original form

robmarch2:
This was based on DTM-49 branch - leaving as a draft onto that branch until it's merged


src/main/java/com/digirati/taxonomy/manager/lookup/model/Word.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public Word(
+            String originalText, int originalStartPosition, int originalEndPosition, String lemma) {
+        this.originalText = originalText;
+        this.originalStartPosition = originalStartPosition;
+        this.originalEndPosition = originalEndPosition;
+        this.lemma = lemma;
+    }
+
+    public Word(CoreLabel coreLabel) {
```
garyttierney: API leakage here.

src/main/java/com/digirati/taxonomy/manager/lookup/TextLookupService.java
```diff
@@ -0,0 +1,72 @@
...
+import com.digirati.taxonomy.manager.lookup.model.LookupResultContext;
+import com.digirati.taxonomy.manager.lookup.model.ConceptMatch;
+import com.digirati.taxonomy.manager.lookup.model.TermMatch;
+import com.digirati.taxonomy.manager.lookup.normalisation.TextNormaliser;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/** Service to orchestrate all the steps in looking up concepts from a piece of input text. */
```
danielgrant: First line of text in Javadoc should start on a new line after the opening tag.

src/main/java/com/digirati/taxonomy/manager/lookup/normalisation/TextNormaliser.java
```diff
@@ -57,24 +60,48 @@ private void initialisePipeline(String languageKey, String languageName) {
...
     }
+
+    /**
+     * Converts a list of {@link Word}s into a single normalised string.
+     *
+     * @param contentWords the {@link Word}s to convert
+     * @return a single normalised string built from the input {@link Word}s
+     */
+    public String normalise(List<Word> contentWords) {
+        StringBuilder normalisedText = new StringBuilder();
```
danielgrant: This entire method can be replaced with:

```
return contentWords.stream().map(Word::getLemma).collect(Collectors.joining(" "));
```

src/main/java/com/digirati/taxonomy/manager/lookup/normalisation/TextNormaliser.java
```diff
@@ -57,24 +60,48 @@ private void initialisePipeline(String languageKey, String languageName) {
...
+
+    /**
+     * Applies the NLP pipeline to a piece of input text to produce a normalised version of it. This
+     * includes tokenisation, sentence splitting, part of speech tagging, and lemmatisation (if this
+     * is supported for the language in question).
+     *
+     * @param text a non-normalised piece of text
+     * @return the normalised form of the input text
+     */
+    public String normalise(String text) {
```
danielgrant: This method only appears to be used by tests - can we get rid of it, or am I missing something?

robmarch2: I've left this in in this PR as a convenience method for tests, but we're going to need it when we get round to normalising the terms going into the Aho-Corasick algorithm

garyttierney: > I've left this in in this PR as a convenience method for tests
We shouldn't be writing production code that only exists to support test code.


> we're going to need it when we get round to normalising the terms going into the Aho-Corasick algorithm

We can add it when we do that.

src/main/java/com/digirati/taxonomy/manager/lookup/TextLookupService.java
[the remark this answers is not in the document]
garyttierney: The style guide disagrees: https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line.

What is the personal preference? I personally prefer folding these comments into one line, to avoid adding vertical padding.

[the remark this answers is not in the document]
danielgrant: Personal preference is to take a newline after the opening tag, but I don't care enough to make a big deal about it. As long as it's as-per the style guide.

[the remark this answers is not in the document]
robmarch2: My personal preference on this would be to have the new line - that's how I wrote it originally but it was removed by formatter so I figured it would be best to leave it without for now

src/main/java/com/digirati/taxonomy/manager/lookup/normalisation/TextNormaliser.java
```diff
@@ -57,24 +60,48 @@ private void initialisePipeline(String languageKey, String languageName) {
...
+        return document.tokens().stream()
                 .filter(this::isContentWord)
-                .forEach(coreLabel -> normalisedText.append(coreLabel.lemma()).append(" "));
-        return normalisedText.toString().trim();
+                .map(Word::new)
+                .collect(Collectors.toList());
     }
 
     private boolean isContentWord(CoreLabel label) {
         return !stopwords.contains(label.originalText()) && !stopwords.contains(label.lemma());
```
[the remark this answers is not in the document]
robmarch2: I've added in stopword normalisation for now - and if, as discussed, we use only pre-normalised stopword lists we'll be able to drop it. Thinking about it again though, I don't think we'll be able to realistically support all languages with static lists (it would be a pain to to raise a PR to add a new list then wait for a review and a deploy every time someone wants to use this with a new language, and some use cases might want more granular control over this than others) so I think we should probably let the user define their own stopword lists, in which case stopword normalisation definitely makes sense.

src/main/java/com/digirati/taxonomy/manager/lookup/normalisation/TextNormaliser.java
```diff
@@ -0,0 +1,125 @@
...
+     * we're using, but language-specific tokenisation should still work
+     *
+     * @param stopwords a collection of common boilerplate words to remove from the text (e.g.
+     *     "the", "a", etc.)
+     * @param languageKey the ISO 639-1 language code for the text to be normalised
+     * @param languageName the name of the language of the text to be normalised
+     */
+    public static Future<TextNormaliser> initialiseNormaliser(
+            Set<String> stopwords, String languageKey, String languageName) {
+        return Executors.newSingleThreadExecutor()
```
danielgrant: I'm not sure I see the point of submitting this task to an executor and returning a `Future`. Obviously the initialisation of the normaliser is a potentially expensive process, and could take some time to complete, but I don't see that alone as being a valid case for using multi-threading.

If the caller gets the `Future<TextNormaliser>` when they call "initialiseNormaliser` or `initialiseEnglishNormaliser`, what are they going to do with it? Answer: they're probably just going to poll it until it returns. Why not return `TextNormaliser` directly and let the call block until it's ready?

Even if the argument is made that the initialisation of the normaliser should happen in another thread, then that the caller should handle the multi-threading. It's not the responsibility of `TextNormaliser` to make assumptions about how the caller wants to handle long running / blocking method calls.

robmarch2: This was in response to one of Gary's comments on the previous change - https://github.com/digirati-co-uk/digirati-taxonomy-manager/pull/6#discussion_r291534298

I don't have a preference either way as long as it's clear to the calling code that this is a long blocking call

PR #13, opened 2019-06-27
Keep engine state in sync with database

robmarch2:


taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/ConceptModel.java
```diff
@@ -39,6 +39,16 @@ public ConceptModel(UUID uuid, Resource resource) {
...
     }
 
+    /**
+     * Gets the UUID of this concept.
+     *
+     * @return the UUID of this concept.
+     * @throws IllegalStateException if the UUID is not present.
+     */
+    public UUID getNullSafeUuid() {
+        return getUuid().orElseThrow(() -> new IllegalStateException("Null UUIDs are not allowed."));
```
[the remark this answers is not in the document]
robmarch2: Yeah, I think we can probably drop the Optional. The very first thing that happens when we create the concept is generating the UUID so it's never going to be null after that. I'll put that into this PR

taxonomy-manager-engine/src/main/java/com/digirati/taxonomy/manager/lookup/ConceptExtractor.java
```diff
@@ -91,12 +90,11 @@ public void updateConcept(
             return;
         }
 
-        originalLabels.forEach(
-                label -> {
-                    Set<UUID> uuids = Sets.newHashSet(conceptLookupTable.get(label));
-                    uuids.remove(conceptUuid);
-                    conceptLookupTable.replaceValues(label, uuids);
-                });
+        originalLabels.forEach(label -> {
```
garyttierney: For simple cases like this (simple case being a stream or collection where we only iterate over, and don't do any pipeline or terminal operations on). I think we should use the imperative for-each instead of streams. It's a little faster and easier to read.

What do you think?

robmarch2: I agree in principle, but I'd rather use either streams or the imperative alternative consistently within a method. Since the stream version is definitely needed for building the set of original labels I went for this form, but I'm happy to pull the building of the set out into its own method so that this can be switched to an imperative for each if you feel strongly about this

garyttierney: I'm not sure the argument of consistency within a method body holds when you're looking at the entire application.

Instead of being inconsistent with the rest of the method body, it's then inconsistent with the rest of the codebase where `Streams` aren't used at all (e.g. `deleteConcept(...)` below, which uses the imperative form and has no `Stream` references). 

I think it's better for an entire codebase to be consistent over consistency between statements in method body. Especially when the rules over code-style within a method are dependent on the code within it.

I don't feel too strongly about this, but it seems like we should be choosing the best tool for the job in every case, and it doesn't feel like we're doing that here.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/Concept.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Gets all labels for this concept for a given language key.
+     *
+     * @param languageKey the language key of the labels to get
+     * @return a set of all labels describing this concept for the given language
+     */
+    default Stream<String> getLabels(String languageKey) {
+        Set<String> labels = new HashSet<>();
+        labels.add(getPreferredLabel().get(languageKey));
```
garyttierney: You can't assume that a preferred label exists for a given language. We assert that a value is present, but not the language associated with it.

taxonomy-manager-engine/src/main/java/com/digirati/taxonomy/manager/lookup/TextLookupService.java
```diff
@@ -24,40 +24,42 @@
...
+
     private final TextNormaliser textNormaliser;
 
     private final ConceptExtractor conceptExtractor;
 
     private final String languageKey;
 
     public static Future<TextLookupService> initialiseLookupService(
-            Stream<ConceptModel> concepts,
+            Stream<? extends Concept> concepts,
```
garyttierney: `Concept` can be something like:

```java
class Term {
    Term(UUID id, Set<String> labels) { ... }
}
```

We don't need to know about concepts other than their identities and labels within the engine.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ConceptExtractionService.java
```diff
@@ -28,22 +30,13 @@
...
-    public void initialiseExtractionService(String languageKey, String languageName) {
-        Stream<ConceptModel> concepts = conceptDao.loadAllRecords().map(record -> {
-            try {
-                return conceptMapper.map(new ConceptDataSet(record));
-            } catch (RdfModelException e) {
-                throw new WebApplicationException(e);
-            }
-        });
+    @PostConstruct
+    public void initialiseExtractionService() {
```
garyttierney: This needs to be done on startup, see `DatabaseInitListener` for an example. You'll want to create a `StartupEvent` listener that injects `TextLookupService`.

```java
class AnalysisInitListener {
    @Inject TextLookupService lookupService; // Or whatever needs initialized at boot.

    void onStartup(@Observes(StartupEvent.class) event) {
         logger.log("Initialized text lookup sevice");
    }
}
```

Since the `IdResolvers` are request scoped you'll get an injection error (there's no `RequestScope` during a `StartupEvent`) so you'll have to change that code to support resolving IDs outside of a request. A simple approach is just adding the base URI of the server as a configuration property instead of pulling it from the request and making the bean `ApplicationScoped`.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ConceptModelRepository.java
```diff
@@ -64,6 +68,7 @@ public ConceptModel create(ConceptModel model) {
 
         var dataset = dataMapper.map(model);
         conceptDao.storeDataSet(dataset);
+        conceptExtractionService.getTextLookupService().addConcept(model);
```
garyttierney: The encapsulating transaction can't complete until the automaton is rebuilt in this case. You need to dispatch an event and handle the concept being added asynchronously instead.

Something like:
```java
@Inject @Stream(“createdConceptModels”)
Emitter<ConceptModel> createdEmitter;

createdEmitter.send(model);
```

See reactive messaging: https://lordofthejars.github.io/quarkus-cheat-sheet/#_version_0_15_0

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ConceptExtractionService.java
robmarch2: Will do.

Btw, I've already resolved the `IdResolver` issue with the `Concept` interface. By explicitly sharing the contract between the `ConceptRecord` and `ConceptModel` to provide what the engine needs there's no need to map between them so the `IdResolvers` aren't needed as part of this any more.

garyttierney: Hm, I don't think bypassing the data model and passing data records straight down is the best solution here, but I don't see any problems it causes. Happy to stick with that.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ConceptModelRepository.java
robmarch2: Unfortunately it looks like there was a regression in Quarkus v0.16.0 which stops the application from even starting when you add the required extension (smallrye-reactive-messaging). It appears to be a race condition between the class loader and one of the logger factories but I've not gotten to the bottom of it yet. I've downgraded locally to 0.15.0 and the application starts correctly but for some reason doesn't inject the repository into the resource. So I'm going to park this until I've resolved the other comments then I'll try and track down exactly what's going wrong.

robmarch2: Here's the stacktrace:

```
14:40:40,286 ERROR [io.qua.dev.DevModeMain] Failed to start quarkus: java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkus.netty.deployment.NettyProcessor#build threw an exception: java.lang.ExceptionInInitializerError
        at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:132)
        at io.quarkus.dev.DevModeMain.doStart(DevModeMain.java:166)
        at io.quarkus.dev.DevModeMain.main(DevModeMain.java:88)
Caused by: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkus.netty.deployment.NettyProcessor#build threw an exception: java.lang.ExceptionInInitializerError
        at io.quarkus.builder.Execution.run(Execution.java:108)
        at io.quarkus.builder.BuildExecutionBuilder.execute(BuildExecutionBuilder.java:121)
        at io.quarkus.deployment.QuarkusAugmentor.run(QuarkusAugmentor.java:104)
        at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:98)
        ... 2 more
Caused by: java.lang.ExceptionInInitializerError
        at io.netty.buffer.AbstractByteBufAllocator.<clinit>(AbstractByteBufAllocator.java:36)
        at io.netty.buffer.Unpooled.<clinit>(Unpooled.java:73)
        at io.netty.handler.codec.http.HttpObjectEncoder.<clinit>(HttpObjectEncoder.java:53)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:315)
        at io.quarkus.netty.deployment.NettyProcessor.build(NettyProcessor.java:42)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at io.quarkus.deployment.ExtensionLoader$1.execute(ExtensionLoader.java:507)
        at io.quarkus.builder.BuildContext.run(BuildContext.java:398)
        at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
        at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2011)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1538)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1395)
        at java.base/java.lang.Thread.run(Thread.java:834)
        at org.jboss.threads.JBossThread.run(JBossThread.java:479)
Caused by: java.lang.NullPointerException: name
        at io.netty.util.internal.logging.AbstractInternalLogger.<init>(AbstractInternalLogger.java:41)
        at io.netty.util.internal.logging.LocationAwareSlf4JLogger.<init>(LocationAwareSlf4JLogger.java:37)
        at io.netty.util.internal.logging.Slf4JLoggerFactory.wrapLogger(Slf4JLoggerFactory.java:54)
        at io.netty.util.internal.logging.Slf4JLoggerFactory.newInstance(Slf4JLoggerFactory.java:49)
        at io.netty.util.internal.logging.InternalLoggerFactory.getInstance(InternalLoggerFactory.java:93)
        at io.netty.util.internal.logging.InternalLoggerFactory.getInstance(InternalLoggerFactory.java:86)
        at io.netty.util.ResourceLeakDetector.<clinit>(ResourceLeakDetector.java:101)
        ... 18 more

```

PR #15, opened 2019-07-02
Fix IndexOutOfBoundsException when engine returns match on first word

robmarch2:


taxonomy-manager-engine/src/main/java/com/digirati/taxonomy/manager/lookup/TextLookupService.java
```diff
@@ -125,6 +125,13 @@ private ConceptMatch createDenormalisedMatch(
         return new ConceptMatch(originalTerm, normalisedMatch.getConceptIds());
     }
 
+    private int getContentWordIndex(String[] previousWords) {
+        if (previousWords.length == 1 && "".equals(previousWords[0])) {
```
garyttierney: The impression I'm getting from hacks like this is that the code around here is quite fragile. I'm going to create a tech debt ticket to re-look at how we're doing this mapping.

garyttierney: DTM-105.

robmarch2: I think whatever we end up doing will be somewhat fragile due to the nature of having to carry round so many different indices but it's worth having a look to see if there's a better way of doing it. My only concern is if we spend time working on this only to shift the same kinds of fragility elsewhere

PR #16, opened 2019-07-04
Create concept autocomplete endpoint

robmarch2:


taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/ConceptDao.java
```diff
@@ -29,6 +29,11 @@ public ConceptDao(DataSource dataSource) {
         return jdbcTemplate.query("SELECT * FROM get_all_concepts()", recordMapper).stream();
     }
 
+    public Stream<ConceptRecord> getConceptsByPartialLabel(String label) {
```
garyttierney: If we already have a full collection, we shouldn't return a `Stream`. Let the API consumer decide if they want a stream or a collection.

If they want to go back to a collection from a `Stream`, they have to copy to a new collection. Going to a `Stream` from a collection is zero-copy.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/autocomplete/ConceptAutocompleteSuggestionMapper.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final String searchTerm;
+
+    ConceptAutocompleteSuggestionMapper(String searchTerm) {
+        this.searchTerm = searchTerm;
+    }
+
+    AutocompleteSuggestion map(ConceptModel conceptModel) {
+        URI uri = conceptModel.getUri();
+        Map<String, String> preferredLabel = conceptModel.getPreferredLabel();
```
garyttierney: Why are we trying to figure out which language matched? We should be only matching against one language when we query the database.

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/taxonomy/ConceptResource.java
```diff
@@ -29,6 +31,12 @@
     @JsonLdFrame(input = "/jsonld/framing/concept.json")
     Response getConcept(@BeanParam ConceptPath params);
 
+    @GET
+    @Path("/by-label/{label}")
+    @Consumes({MediaType.APPLICATION_JSON})
```
garyttierney: This method doesn't consume anything.

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/taxonomy/ConceptResource.java
```diff
@@ -29,6 +31,12 @@
     @JsonLdFrame(input = "/jsonld/framing/concept.json")
     Response getConcept(@BeanParam ConceptPath params);
 
+    @GET
+    @Path("/by-label/{label}")
```
garyttierney: This should be something like `/search`, with a query parameter deciding what the search term is. E.g., ```search?label=whatever```. Then we can extend it in the future without making breaking changes.

taxonomy-manager-rest-server/src/main/resources/db/procedures/skos_concept/R__get_concepts_by_partial_label.sql
```diff
@@ -0,0 +1,12 @@
+DROP FUNCTION IF EXISTS get_concepts_by_partial_label;
+CREATE OR REPLACE FUNCTION get_concepts_by_partial_label(_label text) RETURNS SETOF skos_concept AS
+$$
+BEGIN
+    RETURN QUERY
+        SELECT concept.*
+        FROM skos_concept concept
+        WHERE EXISTS(SELECT 1
+                     FROM jsonb_each_text(concept.preferred_label) AS preferred_label
+                     WHERE preferred_label.value::text ~ ('^' || _label));
```
garyttierney: When we're doing a search it'll be for a specific language, so we need to only interrogate the values associated with that language key (and also the hidden labels, too).

taxonomy-manager-rest-server/src/integrationTest/java/com/digirati/taxman/rest/server/taxonomy/storage/ConceptDaoTests.java
```diff
@@ -98,6 +100,30 @@ public void shouldStoreRelationships() {
...
+        var eleven = new ConceptRecord(UUID.randomUUID());
+        eleven.getPreferredLabel().put("en", "eleven");
+        eleven.getPreferredLabel().put("fr", "onze");
+
+        var dao = new ConceptDao(dataSource);
+        dao.storeDataSet(new ConceptDataSet(one));
+        dao.storeDataSet(new ConceptDataSet(two));
+        dao.storeDataSet(new ConceptDataSet(eleven));
+
+        var expected = Sets.newHashSet(one, eleven);
```
garyttierney: Use Java 11 collection factories instead of Guava ones.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/media/writer/AutocompletionResultWriter.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyWriter;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+public class AutocompletionResultWriter implements MessageBodyWriter<AutocompletionResult> {
```
garyttierney: Why was this needed?

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/taxonomy/ConceptResource.java
garyttierney: cc @stephenwf

garyttierney: > Everyone happy with that or do we want to scope it to a single concept scheme as per the mock api? (edited) 

Probably not scoped to a concept scheme. If you scope it to a concept scheme, there's no way for the UI to let you create cross-concept scheme references.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/autocomplete/ConceptAutocompleteSuggestionMapper.java
robmarch2: So that the frontend can determine which of the preferred labels to display. I disagree that we should only be matching against one language since the point of this is to provide concept IDs to the frontend when the user is trying to build relationships. Since relationships are based on those IDs rather than labels, the language we search for is irrelevant. Why do you think we should only be matching against one language and how do you propose that that frontend gets that language?

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/ConceptDao.java
robmarch2: This was to be consistent with the method above (`'loadAllRecords`). I agree with your point in principle but in this case there's a single consumer of this which does want this as a Stream

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/taxonomy/ConceptResource.java
robmarch2: This was copied from getConcept - happy to remove it from both

garyttierney: Yep, should be removed from there too.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/media/writer/AutocompletionResultWriter.java
robmarch2: This was needed to serialise the `AutocompletionResult` as JSON. Without it there's a runtime error complaining that this doesn't exist

garyttierney: Didn't this work? https://quarkus.io/guides/rest-json-guide

robmarch2: No. Presumably it's because there's already a `MessageBodyWriter` for `RdfModel` so it's trying to use that API for every endpoint

garyttierney: We can get rid of this anyway when we send RDF data back to the client.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/RdfModelBuilder.java
```diff
@@ -59,7 +59,8 @@ public RdfModelBuilder addPlainLiteral(Property property, Map<String, String> va
      * @param embeddedModel The builder used to produce the new embedded model.
      * @throws RdfModelException if the embedded resource was invalid.
      */
-    public RdfModelBuilder addEmbeddedModel(Property property, RdfModelBuilder embeddedModel) throws RdfModelException {
+    public RdfModelBuilder<T> addEmbeddedModel(Property property,
```
garyttierney: Oops, good catch

taxonomy-manager-rest-server/src/integrationTest/java/com/digirati/taxman/rest/server/taxonomy/storage/ConceptDaoTests.java
```diff
@@ -98,6 +98,35 @@ public void shouldStoreRelationships() {
...
+        var dao = new ConceptDao(dataSource);
+        dao.storeDataSet(new ConceptDataSet(one));
+        dao.storeDataSet(new ConceptDataSet(two));
+        dao.storeDataSet(new ConceptDataSet(eleven));
+        dao.storeDataSet(new ConceptDataSet(hundred));
+
+        var expected = List.of(one, hundred);
+        var actual = dao.getConceptsByPartialLabel("on", "en");
+
+        Assertions.assertEquals(expected, actual);
```
garyttierney: I don't think this assertion is comparing what you think. The `equals(...)` contract on `ConceptDataSet` only compares equality of the `uuid` on the 2 records, not the labels. See: https://github.com/digirati-co-uk/taxonomy-manager/blob/master/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/record/ConceptRecord.java#L126

garyttierney: Actually, nevermind. This looks good.

PR #17, opened 2019-07-08
Add implementation of concept extraction API

garyttierney:


taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/analysis/TextAnalyzer.java
```diff
@@ -0,0 +1,58 @@
...
+        var ctx = lookupService.search(input.getText());
+        var matches = ctx.getMatchedConcepts();
+
+        try {
+            var builder = modelFactory.createBuilder(CollectionModel.class);
+            builder.setUri(URI.create("urn:collection"));
+
+            var matchedConcepts = matches.parallelStream()
+                    .flatMap(match -> match.getConceptIds().stream())
+                    .map(concepts::find);
```
robmarch2: I don't think we should make a separate call to get each concept - probably worth adding a db function that takes a set of UUIDs and gives back the relevant concepts.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/CollectionModel.java
```diff
@@ -0,0 +1,26 @@
...
+import com.digirati.taxman.common.rdf.annotation.RdfConstructor;
+import com.digirati.taxman.common.rdf.annotation.RdfContext;
+import com.digirati.taxman.common.rdf.annotation.RdfType;
+import org.apache.jena.rdf.model.Resource;
+import org.apache.jena.vocabulary.DCTerms;
+import org.apache.jena.vocabulary.SKOS;
+
+@RdfType("http://www.w3.org/2004/02/skos/core#Collection")
+@RdfContext({"skos=" + SKOS.uri, "dcterms=" + DCTerms.NS})
+public class CollectionModel implements RdfModel {
```
robmarch2: I've added this class as part of the autocomplete PR, so we should probably make sure we're using the same one in both changes

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/RdfModelBuilder.java
```diff
@@ -52,6 +52,11 @@ public RdfModelBuilder addPlainLiteral(Property property, Map<String, String> va
         return this;
     }
 
+    public RdfModelBuilder addEmbeddedModel(Property property, RdfModel model) throws RdfModelException {
```
robmarch2: This should return a `RdfModelBuilder<T>`

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/lifecycle/AnalyzerInitListener.java
```diff
@@ -0,0 +1,23 @@
...
+
+@ApplicationScoped
+public class AnalyzerInitListener {
+    private static final Log logger = LogFactory.getLog(AnalyzerInitListener.class);
+
+    @Inject
+    TextLookupService service;
+
+    void onStartup(@Observes StartupEvent event) {
+        logger.info("Finished building TextLookupService");
```
robmarch2: How do we know this listener fires after the `TextLookupService` has been built?

garyttierney: `TextLookupService` is injected into this class, so it can't be constructed until `TextLookupService` is initialized.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/RdfModelBuilder.java
garyttierney: Yep, I'll leave the other one untyped to make it conflict with your work a bit less.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/analysis/TextAnalyzer.java
garyttierney: This one is a little odd. It doesn't seem to be possible to get PostgreSQL to use the unique UUID index when we're querying for a collection of concepts (here's the plan: http://tatiyants.com/pev/#/plans/plan_1562595043043).

Instead, it does a sequential scan. I suppose this is down to 1 of 2 reasons: 

1) the query planner is unable to optimize the `ANY(_uuids)` condition. 
2) the query planner can optimize it and decides a sequential scan is faster.

At a large number of concepts that sequential scan starts to become a problem (especially with wide tables containing JSON), but it may turn out that index scans start kicking in at a high enough volume. I'll push an implementation that uses the array approach and we can go with that until it becomes an issue.

PR #19, opened 2019-07-08
Authenticate all management endpoints via JWT

robmarch2:


taxonomy-manager-rest-server/src/test/resources/jwtClaims.json
```diff
@@ -0,0 +1,12 @@
+{
+  "iss": "https://dev.taxman.digirati.com/",
+  "jti": "a-123",
+  "sub": "jdoe-using-jwt-rbac",
+  "upn": "jdoe@quarkus.io",
+  "preferred_username": "jdoe",
+  "aud": "using-jwt-rbac",
```
garyttierney: Hm, if the `aud` field is present doesn't match the server the token was sent to (via the `Authorization` header) the implementation is supposed to throw an error. Does Quarkus not complain about this?

robmarch2: No, the only ones it seems to care about are `iss` and `groups` atm. Possibly needs a bit of config to make it check that too

garyttierney: That makes it non-compliant with RFC 7519, which isn't a huge deal but definitely something to keep in mind. I'll have a look over at the SmallRye JWT implementation.

taxonomy-manager-rest-server/.gitignore
```diff
@@ -1,2 +1,3 @@
 /build
 /out
+/src/test/resources/privateKey.pem
```
garyttierney: Why are we ignoring a fake generated private key?

robmarch2: Because it's the real one. We need the real public key checked into git for use in the Application.properties, so we need to generate the test JWT using the real private key if we want to verify that it's letting the right people in

danielgrant: I'm still not entirely clear on why this is being ignored - as @garyttierney mentioned, it's a fake private key (i.e. generated solely for the purpose of running tests). It's not used for anything outside of this.

Also, if we aren't checking it in, how does Jenkins execute the JWT tests without it?

taxonomy-manager-rest-server/src/main/resources/application.properties
```diff
@@ -10,6 +10,11 @@ quarkus.flyway.connect-retries=10
 quarkus.flyway.table=flyway_quarkus_history
 quarkus.flyway.locations=db
 
+quarkus.smallrye-jwt.enabled=true
+quarkus.smallrye-jwt.auth-mechanism=MP-JWT
+mp.jwt.verify.publickey.location=META-INF/publicTestKey.pem
+mp.jwt.verify.issuer=https://dev.taxman.digirati.com/
```
danielgrant: We need to come up with a better way of managing this - the environment (i.e. `dev`) can be passed at runtime, perhaps as an environmental variable. But hardcoding it here is no use, this will break as soon as we deploy to any environment other than `dev`.

In fact, we shouldn't be hardcoding this value at all, including the hostname.

taxonomy-manager-rest-server/src/test/resources/jwtClaims.json
```diff
@@ -0,0 +1,12 @@
+{
+  "iss": "https://dev.taxman.digirati.com/",
```
danielgrant: Can we use something other than `dev.taxman.digirati.com` here for the purpose of running tests please.

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/Roles.java
```diff
@@ -0,0 +1,8 @@
+package com.digirati.taxman.rest;
+
+public final class Roles {
+
+    public static final String TAXMAN_ADMIN = "taxman-admin";
```
danielgrant: Two points:
1. Typically I've seen parts of a role name separated by underscores rather than hyphens, I think we should do the same here; and
2. Any reason why this isn't an enum instead?

danielgrant: Also, let's drop the `TAXMAN` part - we haven't branded the product yet so we should avoid all colloquial terminology we use in house.

taxonomy-manager-rest-server/.gitignore
robmarch2: Ah, sorry - there was some face-to-face discussion on this point that didn't make it onto here. I included the key in the second commit of this PR but pre-commit has a rule of failing the build if the change contains a private key so we backed it out again. As for how to run tests end-to-end, we don't have anything that needs this yet (there's a ticket in the backlog to work out how to do it) - this change only needs that private key for running tests manually. Now that I think about it again though, I think it would probably be better to check this key in, and drop the pre-commit rule that makes it an issue, to save having to manually pass this key around

stephenwf: Is it bad practice to generate the key in CI?

danielgrant: @stephenwf no, and in fact @robmarch2 I think that would probably be a better idea - it feels wrong to check any key (public or private) into GitHub regardless of what it's used for. It's trivial to generate a temporary key during CI.

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/Roles.java
robmarch2: Happy to make the changes to the role name.

The reason I went with String constants over an enum is that the calling code needs a String, so if this was an enum the calling code (which is part of a Quarkus extension) wouldn't be restricted to the enum values. Given that the only thing this would need to do if it was an enum is provide a String, making it a class with String constants is a better way of doing it in this case.

taxonomy-manager-rest-server/src/main/resources/application.properties
robmarch2: I hardcoded this with the intention of having the deploy overwrite this, but it looks like quarkus supports having environment variables in here natively so I think that would better

PR #20, opened 2019-07-09
Add implementation of recursive relations endpoint

garyttierney:
Some notes.

* Max depth of 5 -- this is to prevent unbounded graph walks.
* Detection of cyclic references -- goes without saying.
* No validation on input -- endpoint will 500 if you give it in invalid type.

### DIFF
--- a/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptResource.java
+++ b/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptResource.java
@@ -1,8 +1,12 @@
 package com.digirati.taxman.rest.server;
 
+import com.digirati.taxman.common.taxonomy.CollectionModel;
 import com.digirati.taxman.common.taxonomy.ConceptModel;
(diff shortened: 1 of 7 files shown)


garyttierney:
Oops, bit of a mistake made here. The function is supposed to be returning `skos_concept` records, since this is exposed as a flat list rather than relations on a single concept. Updated the query to reflect that.

garyttierney:
> Client has to know the relationship name, correct?

That's right. It's case-insensitive and should accept any semantic relation value it's given (e.g., broader, narrower).

> All looks good. The query is scary, but I really like it - it offloads the worrying about recursive data collection to the DB, so I assume it will be as fast as it can be under the circumstances. Clever depth/cycle protection.

The only remaining optimization to be made is the fact that the final part of the query:

```sql
        SELECT sct.*
        FROM relationships r
        INNER JOIN skos_concept sct
            ON sct.id = r.target_id;
```

Is a sequential scan. Probably just needs time to build data up and then a `VACUUM ANALYZE`.

> I am not familiar with syntax greatly, but in the query lack of labels makes it a bit hard to read (to remember that this boolean expression will become the cycle field.

I've added some column aliases.

PR #22, opened 2019-07-11
Ensure duplicate Scheme-Concept links are not created

robmarch2:
Previously if a concept scheme had top concepts A and B, then an update was pushed to remove A and add C, a duplicate entry was created for B. This was because we were relying on a conflict resolution strategy to prevent the duplicate from being added despite there not being any constraints that could trigger a conflict. This change prevents the duplicate from being added by overwriting the list of associated concepts instead of attempting to detect and ignore duplicates


garyttierney:
Ah, there's a missing index. If you add a unique index over `concept_scheme_concept(concept_scheme_id, concept_id, is_top_concept)` the conflict resolution will work.

We want to try and avoid database rewrites (i.e. DELETE then re-INSERT) or the on disk database files will grow until we run a `VACUUM` manually.

PR #24, opened 2019-07-12
Create project endpoints

robmarch2:


taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerProjectResource.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public Response getProject(ProjectPath projectPath) {
+        return Response.ok(projectModelRepository.find(projectPath.getProjectSlug())).build();
+    }
+
+    @Override
+    public Response updateProject(ProjectPath projectPath, @Valid ProjectModel project) {
+        projectModelRepository.update(project);
+        return Response.noContent().build();
```
danielgrant: I'm concerned that `204 No Content` is not the correct response to return here. Typically I've seen this used in response to a `DELETE`, where the target resource has been deleted and thus there's nothing to return.

For any kind of update call (e.g. `POST`, `PUT`) I think it makes sense to return _the updated object_ a response to that operation, which would imply `200 OK`. This allows for the client to immediately update their view of that resource (e.g. in the browser) without having to make additional calls.

I understand that there may be other similar calls outside of the project operations where we're returning `204 No Content` in response to an update operation, so this may be a larger issue.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ProjectModelRepository.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Stores a new project in the database, along with mappings between the project and any associated concept schemes.
+     *
+     * @param project the project to create
+     * @return the created project
+     */
+    @Transactional(Transactional.TxType.REQUIRED)
+    public ProjectModel create(ProjectModel project) {
+        ProjectDataSet dataSet = projectMapper.map(project);
+        projectDao.storeDataSet(dataSet);
```
danielgrant: This might not be the correct place to handle this, but are we doing any checks to ensure that a project with the provided name doesn't already exist, before we get this far? I can't see where such checks are happening...

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/ProjectDao.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Retrieves a project with a given slug identifier.
+     *
+     * @param slug the identifier of the project to retrieve
+     * @return the project with the given slug
+     */
+    public ProjectDataSet loadDataSet(String slug) {
+        Object[] recordArgs = {slug};
+        int[] recordTypes = {Types.OTHER};
```
danielgrant: Is this actually a `Types.OTHER`? Looks like a `Types.VARCHAR` to me...

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ProjectModelRepository.java
robmarch2: We're not - if the slug already exists it'll store it as an update to the existing project, which now that I think about it again is wrong. I'll add something here to check that

danielgrant: > if the slug already exists it'll store it as an update to the existing project

Project creation comes in as a `POST`, update comes in as a `PUT`. You can't store it as an update if someone `POST`'s to the `/projects` endpoint using an existing slug, that can only occur if someone `PUT`'s to the `/project/{slug}` endpoint.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerProjectResource.java
robmarch2: This is the case for the updates on concepts and concept schemes as well so I'd prefer to tackle them all in one go. I'm not sure I agree that returning the modified object would be useful to the client though, since in all three cases we're just storing what they told us to store rather than doing any server side processing on that data

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ProjectModelRepository.java
robmarch2: The DAO uses the same method in either case. The only difference between the two endpoints once you get into them is what we return from them

robmarch2:
> There's an overall lack of condition / error handling. For example, there's nothing that tests whether a given project slug already exists during creation, instead it appears we're depending on an error bubbling up from the DAO. Instead I would expect something like a `409 Conflict` to be returned.

This should definitely be fixed as part of this change

> Similarly, there's no integrity checking in the `PUT` operation. What happens if two clients retrieve the same project, and attempt to update it one after the other? The second clients update will overwrite the first clients changes, because the second client has no way of determining that a modification has been made to the project between his `GET` and `PUT`. A typical way of resolving this is using an ETag header, that is stored alongside the record on the server and returned in the `GET` and provided in the `PUT`. This value changes on every update to the record, so the server can determine if a clients provided ETag value is current or not when their `PUT` comes in, and if not can return e.g. `304 Not Modified`. I suspect this issue might exist across the board though, for all other operations, but I would consider it an essential feature or we're going to get clients stepping on each others toes.

This same issue will affect all endpoints that persist something across the entire application. We should definitely fix it but since it's already so widespread I'd prefer if that was done for everything in a separate change

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ProjectModelRepository.java
danielgrant: That's even more wrong, it just introduces unpredictable behaviour. `POST`'ing to `/projects` should **not** update the project if it already exists, only `PUT` to `/projects/{slug}` should do that.

The behaviour exposed by each endpoint must be tightly controlled, it shouldn't (read: mustn't) leak like that.

robmarch2: This same issue affects concept schemes too. I'll fix it for projects as part of this change then make another one fixing it for concept schemes

robmarch2: Actually, scratch that. This exact issue doesn't affect concept schemes because we generate the UUID as part of the creation, although a related issue exists whereby the `PUT` can create a new scheme. The fix to that should be to 404 if the scheme being updated doesn't exist - which requires the new `EmptyResultExceptionMapper` class from this change plus a single line added to the `ConceptSchemeModelRepository` - so I'll fix that as part of this change

PR #25, opened 2019-07-12
Docker compose - development

stephenwf:
Fleshes out the docker-compose to include the JVM docker file and sample configuration for running locally. Useful for non-java developers getting the service up and running for integration or testing. By default will run on `localhost:8104`.

Can parameterise the docker-compose a little more to accept environment variables instead, add in something like an `.env.dist` example etc. Let me know!


docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+      POSTGRES_DATABASE: taxman
+      POSTGRES_PASSWORD: taxman
+  taxman:
+    build:
+      dockerfile: ./dockerfiles/Dockerfile.jvm
+      context: .
+    links:
+      - postgres
+    ports:
+      - 8104:8080
```
danielgrant: Port `8104`? The backend will bind to `8080` on the container by default.

README.md
```diff
@@ -11,6 +11,21 @@ To build the Digirati Taxonomy Manager, use the included gradle wrapper:
 ./gradlew build
 ```
 
+Once you have built the project locally you can use docker-compose to get an environment up, with a database and sample configuration:
+```
```
danielgrant: [MD031](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines), there should be blank lines around fences.

Same for the other fences that have been added which don't have blank lines.

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: '3'
```
danielgrant: Any particular reason why you took the `.devel` off of this? It was there to make it explicitly clear that this isn't a production ready Docker Compose configuration.

docker-compose.yml
stephenwf: No problem, updated that. Force of habit choosing random ports, always run into conflicts.

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+services:
+  postgres:
+    image: postgres
```
danielgrant: Can we pin to a specific version here please, it's bad practice to let this default to `latest`.

docker-compose.yml
stephenwf: To improve the developer experience. 
```
docker-compose -f docker-compose.devel.yml up
```

vs.
```
docker-compose up
```

I can add a comment to the readme to clarify that it's not a production ready instance. Although at this point, adding some parameterisation and it could be production ready for micro instances.

danielgrant: :thinking: @robmarch2 @garyttierney thoughts?

robmarch2: I don't have a particularly strong opinion in this instance seeing as I only ever tend to run the application via gradle. In general I tend to come down on the side of being explicit, but in this case if this is the same `docker-compose.yml` that will be used when it is prod ready, I don't see much point in temporarily having the `.devel`. If on the other hand, the intention is to have separate ones for local development vs prod, it should stay.

PR #26, opened 2019-07-15
Ensure paths to JSON-LD frames can be read when running via docker

robmarch2:
Previously the path to the JSON-LD frames were being interpreted as relative when running the application via gradle but as absolute when running via docker. This was causing a null pointer due to the fact that no such file with the given absolute path existed. The fix is to make all the paths to the JSON-LD frames explicitly relative.


robmarch2:
> That's a good catch. I trust that you've tested that and the solution indeed behaves as expected now.

I have indeed, although in the course of testing I've found another issue that only occurs when you run it via docker so I'm going to add it into this PR. When adding a concept, it doesn't get picked up by the analysis engine unless a second concept is also added before running the analysis

PR #32, opened 2019-07-29
Implemented enhanced Jenkins pipeline for CI / CD

danielgrant:
This pipeline builds upon the existing pipeline by adding the following:

- All pushes to the `master` branch now result in a release candidate tag being automatically created, along with a corresponding release in GitHub, and deployment to `DEV`.
- Running a release tag through the pipeline results in creation of the corresponding release in GitHub, and deployment to `DEV`.
- All other branches (e.g. feature, bugfix, chore, etc) result in the usual set of test and build steps, but with no tagging, release creation or deployment taking place.

Branches, PR's and tag builds can be re-run at any time via the Jenkins UI. Any version of the backend can be deployed to any environment via the `taxonomy-manager` build job, by running the job with the relevant parameters.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -214,12 +214,12 @@ pipeline {
             }
 
             steps {
-                build job: "$DEPLOYMENT_JOB",
-                      parameters:  [
-                          [$class: 'StringParameterValue', name: 'ENVIRONMENT', value: "$DEPLOYMENT_ENV"],
-                          [$class: 'StringParameterValue', name: 'BACKEND_IMAGE_TAG', value: "${tagVersion}"]
-                      ],
-                      propagate: true
+                build job: DEPLOYMENT_JOB,
+                    parameters:  [
+                        stringParam(name: 'ENVIRONMENT', value: DEPLOYMENT_ENV),
+                        stringParam(name: 'FRONTEND_VERSION', value: tagVersion)
+                    ],
+                    propagate: true
             }
         }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -217,7 +217,7 @@ pipeline {
                 build job: DEPLOYMENT_JOB,
                     parameters:  [
                         stringParam(name: 'ENVIRONMENT', value: DEPLOYMENT_ENV),
-                        stringParam(name: 'FRONTEND_VERSION', value: tagVersion)
+                        stringParam(name: 'BACKEND_IMAGE_TAG', value: tagVersion)
                     ],
                     propagate: true
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -128,12 +128,7 @@ pipeline {
 
         stage('build image') {
             steps {
-                script {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -106,44 +132,94 @@ pipeline {
...
-                build job: "$DEPLOYMENT_JOB",
-                      parameters:  [
-                          [$class: 'StringParameterValue', name: 'ENVIRONMENT', value: "$DEPLOYMENT_ENV"],
-                          [$class: 'StringParameterValue', name: 'BACKEND_IMAGE_TAG', value: "${version}"]
-                      ],
-                      propagate: true
+                build job: DEPLOYMENT_JOB,
+                    parameters:  [
+                        stringParam(name: 'ENVIRONMENT', value: DEPLOYMENT_ENV),
+                        stringParam(name: 'FRONTEND_VERSION', value: tagVersion)
```
garyttierney: Should this not be `BACKEND_VERSION`?

danielgrant: :man_facepalming:

Jenkinsfile
```diff
@@ -106,44 +132,94 @@ pipeline {
...
+                anyOf {
+                    branch 'master'
+                    tag pattern: RELEASE_TAG_REGEX, comparator: 'REGEXP'
+                }
+            }
+
+            steps {
+                withCredentials([usernamePassword(credentialsId: "$IMAGE_CREDS_JENKINS_ID", usernameVariable: 'IMAGE_REGISTRY_USERNAME', passwordVariable: 'IMAGE_REGISTRY_PASSWORD')]) {
+                    sh 'docker login $IMAGE_REGISTRY --username $IMAGE_REGISTRY_USERNAME --password $IMAGE_REGISTRY_PASSWORD'
+                    sh "docker tag \$IMAGE_REPOSITORY:latest \$IMAGE_REGISTRY/\$IMAGE_REPOSITORY:${tagVersion}"
```
garyttierney: Is it possible to use the Docker API instead here?

garyttierney: The Jenkins Docker API, i.e., the `docker.image(...).push()` stuff.

Jenkinsfile
danielgrant: Fixed.

Jenkinsfile
danielgrant: Done :+1:

danielgrant: Alas, due to this bug: https://issues.jenkins-ci.org/browse/JENKINS-44609 it's not possible with `Dockerfile.jvm` in its current form.

Until that bug is resolved, there are only two options:

1. Continue to use `sh 'docker...`; or
2. Refactor `Dockerfile.jvm` to remove the `AS <aslias>` parts from the `FROM` lines, and reference the different stages by index (e.g. `COPY --from=0 ...`).

danielgrant:
> Mostly just the same as the frontend release. We should move the common step blocks into a shared library at some point.

Agreed.

PR #38, opened 2019-10-08
DTM-142: Symmetric Relationships

p-kaczynski:
1. When updating a concept, ensure that the narrower/broader also have symmetric reciprocal relations to the concept (code update)
2. When updating a concept's relations, delete the relations not present in incoming data (sproc update)

### DIFF
--- a/taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/ConceptRelationshipType.java
+++ b/taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/ConceptRelationshipType.java
@@ -49,4 +49,17 @@ public enum ConceptRelationshipType {
 
         return transitive ? transitiveRelationshipProperty : relationshipProperty;
     }
+
+    public ConceptRelationshipType inverse() {
+        switch (this) {
+            case BROADER:
+                return NARROWER;
+            case NARROWER:
+                return BROADER;
+            case RELATED:
+                return RELATED;
+            default:
+                return null; // Or throw...
+        }
+    }
 }
--- a/taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/ConceptRelationshipType.java
+++ b/taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/ConceptRelationshipType.java
@@ -59,7 +59,7 @@ public enum ConceptRelationshipType {
             case RELATED:
                 return RELATED;
             default:
-                return null; // Or throw...
+                throw new IllegalStateException(String.format("ConceptRelationshipType of type '%s' cannot be inverted.", this.name()));
         }
     }
 }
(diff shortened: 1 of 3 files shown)


taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ConceptModelRepository.java
```diff
@@ -86,10 +91,59 @@ public void update(ConceptModel model) {
...
 
+    /**
+     * Analyze the changes between relationships and create symmetric relations if needed
+     * @param model new model
+     */
+    private void applySymmetricRelationChanges(ConceptModel model) {
+
+        final boolean transitive = false;
+
+        BiFunction<ConceptModel, ConceptRelationshipType, Set<UUID>> getRelationshipsOrEmpty
```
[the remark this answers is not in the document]
p-kaczynski: > 
> 
> This doesn't capture anything relevant from the scope of `applySymmetricRelationChanges`, so there's no reason for this to be a lazily initialized closure. It should be moved to a method instead.
> 
> Maybe:
> 
> ```java
> class ConceptModel {
>     public Stream<UUID> getRelationshipUuids(ConceptRelationshipType type, boolean transitive, ConceptIdResolver resolver) {
>         // ...
>     }
> }
> ```

Yeah, it did and then it didn't but stayed as it is ;) Will change.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/taxonomy/ConceptRelationshipType.java
```diff
@@ -49,4 +49,17 @@ public Property getSkosProperty(boolean transitive) {
...
+    public ConceptRelationshipType inverse() {
+        switch (this) {
+            case BROADER:
+                return NARROWER;
+            case NARROWER:
+                return BROADER;
+            case RELATED:
+                return RELATED;
+            default:
+                return null; // Or throw...
```
garyttierney: Let's throw an `IllegalStateException` here telling the user that `this` can't be inversed

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/ConceptModelRepository.java
```diff
@@ -115,4 +168,40 @@ public ConceptModel create(ConceptModel model) {
...
     }
+
+    /**
+     * Analyze the changes between relationships and create symmetric relations if needed
+     *
+     * @param model The new data for the Concept
+     * @param existing (Can be null) The current version of the Concept, if one exists
+     */
+    private void applySymmetricRelationChanges(ConceptModel model, ConceptModel existing) {
+
```
garyttierney: Unnecessary whitespace here, also can we run a formatter over this code? There's inconsistent spacing between characters, and arguments should be on the same line as the parameters:
```
BiConsumer<UUID, ConceptRelationshipType> createRelationship =
         (relatedUuid, relationshipType) -> {
         };
```

taxonomy-manager-rest-server/src/main/resources/db/procedures/skos_concept_semantic_relation/R__update_concept_semantic_relations.sql
```diff
@@ -12,17 +12,45 @@ FROM jsonb_array_elements(_relations) data
 WHERE sc.uuid IS NULL
 ON CONFLICT DO NOTHING;
 
--- First, add all new entries to the database, ignoring items that are already there.
+-- Delete where relations do not exist in provided JSON data
```
garyttierney: Can we re-format the DELETE part here

PR #39, opened 2019-10-23
Add test verifying that source URIs are preserved

garyttierney:


taxonomy-manager-rest-server/src/test/java/com/digirati/taxman/rest/server/ServerConceptResourceTest.java
```diff
@@ -22,12 +28,42 @@ public void createConcept_JsonLd() {
...
+                        .body("'dcterms:source'.'@id'", equalTo("https://example.org/my-id-001"))
+                    .and()
+                    .extract()
+                        .header("Location");
+
+        givenJsonLdRequest(getClass(), "concept--update-with-source.json", Map.of("@id", conceptLocation))
+                .when()
+                    .put(conceptLocation)
+                .then()
+                    .statusCode(204);
```
p-kaczynski: Technically this is updating pre-existing `dcterms:source` with a new one rather than sending the existing non-default and expecting the Concept to have unchanged `dcterms:source`...

...but this is the same thing essentially, so this is a more robust test (allowing FE to put desired value, useful for fixing broken stuff I guess.)

garyttierney: On the other side of this: sending a `Concept` resource to taxman without a `dcterms:source` property is telling the backend you've updated the concept, and removed the `dcterms:source` field. `PUT`s are always full updates.

Whether or not that is what we want in all cases is still open for discussion, but the general model of idempotent updates seems to be working so far.

taxonomy-manager-rest-server/src/test/java/com/digirati/taxman/rest/server/testing/util/RestAssuredUtils.java
```diff
@@ -3,16 +3,47 @@
...
+        try (InputStream resourceAsStream = caller.getResourceAsStream(body)) {
+            json = new String(resourceAsStream.readAllBytes(), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to open request body identified by " + body);
+        }
+
+        var base = Json.createReader(new StringReader(json)).readObject();
+        var merged = Json.createObjectBuilder(base);
+
+        mergeProperties.forEach(merged::add);
```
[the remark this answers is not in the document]
garyttierney: Yeah, taxman gets the UUID from the `@id` property, so needed that injected in here somewhere. Might be a cleaner way of re-assembling this JSON, or maybe a better way of resolving UUIDs from requests in the backend.

PR #43, opened 2019-11-12
DTM-113: Deletion of Concepts, Concept Schemes and Projects

p-kaczynski:
## Goal
1. Allow making DELETE REST calls to Concept, Concept Scheme and Project resources
1. Soft delete the aforementioned resources

### Additional Goals
1. Ensure that deletion of Concepts affects the extraction

### API
Cookie-cutter endpoint dispatching request to the relevant repository class. Returns 204 No Content as per guidelines.

#### Projects and Concept Schemes
Direct pass-through to the DAO layer
#### Concepts
Call DAO layer, but also create and dispatch a `ConceptEvent`.

### DAO Layer
Execute the stored procedure that will perform operations necessary for deletion of the resource.

*Note:* using `jdbcTemplate.update` as procedures are simple and other ways to call seem unnecessarily convoluted. This way we have more consistent code with no obvious downsides due to that choice. 

#### DDL Changes
1. Added `deleted boolean default false not null` column to `skos_concept`, `skos_concept_scheme` and `project` tables
1. Created an index based on the deletion status on each table
1. Created a `<table_name>_ex` (`ex`-isting) view for each table

#### Stored Procedures
1. Modified all procedures where data was read from any of the three modified tables, to use the `_ex` view instead
  1. This is a very simple and easy win, as it abstracts the deletion status from the codebase
1. Added `delete_<table_name>(unique_identifier)` procedures, which simply switch the `deleted` column value to `true`

## Event System
As of the current base branch version the Event System is not being used.

To ensure that the deletion of a Concept affects the Extraction without a need for a restart (which is a big hurdle in a PRD environment), I have modified and built upon the existing Event System infrastructure.

### ConceptEvent
1. Added new `DELETED` type
1. Added new `ConceptModel previous` field
  1. Rationale:
      1. `CREATED` event: no previous -> nothing to delete. New item -> things to index.
      1. `DELETED` event: previous -> delete from index. No new item -> nothing to add.
      1. `UPDATED` event: prvious -> delete from index. New item -> things to index.
      1. As can be seen, this setup allows very simple and not repetitive code to handle all three cases with just two methods.
1. Added appropriate constructor and aligned code to the changes

### ConceptEventListener
1. Subscribed the existing Listener to the events of type `ConceptEvent`.
1. Depending on the existence of previous and/or new item in the event, the old ones are unindexed and the new ones are indexed.

In the end, those changes also implement previously unsupported stories:
* When a new Concept is added to the Taxonomy, it can be immediately returned, if applicable, from Extraction endpoint
* When a Concept is modified (e.g. new Alt Label is added), the changes are reflected in the Extraction endpoint immediately.

As well as the desired "deletion" one.

### Frontend
* Project Deletion button in the UI does not make the DELETE call to the backend.
### Backend
* Project Deletion action does not affect the Extraction. This is due to the lack of Project Scoping. When implementing Project Scoping, changes must be made to also treat the deleted projects correctly.

### DIFF
--- a/taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
+++ b/taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
@@ -1,0 +1,53 @@
+package com.digirati.taxman.common.rdf.identity;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.regex.Pattern;
+
+public class IdResolver {
+
+    // This does not need to be ever instanciated
+    private IdResolver() {
+    }
+
+    public static Optional<UUID> resolve(String uri, Pattern pattern) {
+        if (uri == null)
+            return Optional.empty();
+
+        try {
+            var parsedUri = new URI(uri);
+            return resolve(parsedUri, pattern);
+        } catch (URISyntaxException ignored) {
+            return Optional.empty();
+        }
+    }
+    public static Optional<UUID> resolve(URI uri, Pattern pattern) {
+        if (uri == null || pattern == null) {
+            return Optional.empty();
+        }
+
+        if (uri.getPath() == null) {
+            return Optional.empty();
+        }
+
+        var match = pattern.matcher(uri.getPath());
+        if (!match.find()) {
+            return Optional.empty();
+        }
+
+        var uuid = match.group(1);
+        return Optional.of(UUID.fromString(uuid));
+    }
+
+//    public static URI resolve(UUID uuid, URI uri, String template) {
+//
+//        return UriBuilder.fromUri(template)
+//                .scheme(uri.getScheme())
+//                .host(uri.getHost())
+//                .port(uri.getPort())
+//                .resolveTemplate("id", uuid.toString())
+//                .build();
+//    }
+}
--- a/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
+++ b/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
@@ -32,7 +32,11 @@ public class ServerConceptSchemeResource implements ConceptSchemeResource {
     public Response getConceptScheme(ConceptSchemePath params) {
         var model = conceptSchemes.find(params.getUuid());
 
-        return Response.ok(model).build();
+        if (model.isPresent()) {
+            return Response.ok(model.get()).build();
+        } else {
+            return Response.status(Response.Status.NOT_FOUND).build();
+        }
     }
 
     @Override
--- a/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
+++ b/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
@@ -42,4 +42,11 @@ public class ServerConceptSchemeResource implements ConceptSchemeResource {
 
         return Response.noContent().build();
     }
+
+    @Override
+    public Response deleteConceptScheme(ConceptSchemePath params) {
+        conceptSchemes.delete(params.getUuid());
+
+        return Response.noContent().build();
+    }
 }
--- a/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/event/ConceptEventListener.java
+++ b/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/event/ConceptEventListener.java
@@ -25,23 +25,25 @@ public class ConceptEventListener {
     TermIndex<UUID> index;
 
     @Subscribe
-    public void ConceptChangeEvent(ConceptEvent conceptEvent) {
-        if (conceptEvent.getPrevious() != null)
-            Remove(conceptEvent.getPrevious());
+    public void conceptChangeEvent(ConceptEvent conceptEvent) {
+        if (conceptEvent.getPrevious() != null) {
+            remove(conceptEvent.getPrevious());
+        }
 
-        if (conceptEvent.getConcept() != null)
-            Add(conceptEvent.getConcept());
+        if (conceptEvent.getConcept() != null) {
+            add(conceptEvent.getConcept());
+        }
     }
 
-    private void Add(ConceptModel concept) {
-        ConsumeLabels(concept, index::add);
+    private void add(ConceptModel concept) {
+        consumeLabels(concept, index::add);
     }
 
-    private void Remove(ConceptModel previous) {
-        ConsumeLabels(previous, index::remove);
+    private void remove(ConceptModel previous) {
+        consumeLabels(previous, index::remove);
     }
 
-    private void ConsumeLabels(ConceptModel concept, BiConsumer<UUID, String> consumer) {
+    private void consumeLabels(ConceptModel concept, BiConsumer<UUID, String> consumer) {
         var labelExtractor = new ConceptLabelExtractor(concept);
 
         labelExtractor.extractTo((property, literal) -> {
--- a/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/event/ConceptEventListener.java
+++ b/taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/event/ConceptEventListener.java
@@ -2,6 +2,7 @@ package com.digirati.taxman.rest.server.infrastructure.event;
 
 import com.digirati.taxman.analysis.index.TermIndex;
 import com.digirati.taxman.common.taxonomy.ConceptLabelExtractor;
+import com.digirati.taxman.common.taxonomy.ConceptModel;
 import com.google.common.eventbus.Subscribe;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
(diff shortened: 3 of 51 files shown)


taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
```diff
@@ -42,4 +42,11 @@ public Response updateConceptScheme(ConceptSchemePath params, @Valid ConceptSche
 
         return Response.noContent().build();
     }
+
+    @Override
+    public Response deleteConceptScheme(ConceptSchemePath params) {
+        conceptSchemes.delete(params.getUuid());
+
+        return Response.noContent().build();
```
garyttierney: Should be 410 Gone.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/event/ConceptEventListener.java
```diff
@@ -1,18 +1,52 @@
...
 public class ConceptEventListener {
 
+    @ConfigProperty(name = "taxman.analysis.default-lang.key", defaultValue = "en")
+    String defaultLanguageKey;
+
     @Inject
     TermIndex<UUID> index;
 
+    @Subscribe
+    public void ConceptChangeEvent(ConceptEvent conceptEvent) {
```
garyttierney: Style  should be consistent with the rest of the project (braces on even 1-statement branches, camelCase members, PascalCase nominals).

taxonomy-manager-rest-server/src/main/resources/db/migration/V9__soft_delete.sql
```diff
@@ -0,0 +1,68 @@
+-- CONCEPT
+
+alter table skos_concept
+    add deleted boolean default false not null;
+
+create index skos_concept_note_deleted
+    on skos_concept (deleted)
```
garyttierney: Hm, not sure what this index is for. We create an index over the 2 possible values of `deleted`: [0,1]. If we filter out the `deleted` records, this index only stores records that aren't deleted. Since `deleted` is the only key though, there's no range scans or quick lookups we get from this.

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/taxonomy/ProjectResource.java
```diff
@@ -3,17 +3,9 @@
 import com.digirati.taxman.common.rdf.annotation.jsonld.JsonLdFrame;
 import com.digirati.taxman.common.taxonomy.ProjectModel;
 import com.digirati.taxman.rest.MediaTypes;
-import com.digirati.taxman.rest.Roles;
 
-import javax.annotation.security.RolesAllowed;
 import javax.validation.Valid;
-import javax.ws.rs.BeanParam;
```
garyttierney: Wildcard imports should be avoided. Would be better to revert back.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
p-kaczynski: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

Should it though?

taxonomy-manager-rest/src/main/java/com/digirati/taxman/rest/taxonomy/ProjectResource.java
p-kaczynski: I blame IntelliJ. Can revert.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
garyttierney: Ah, sorry, you're right. 410 Gone is what we should send if someone requested a deleted record.

taxonomy-manager-rest-server/src/main/resources/db/migration/V9__soft_delete.sql
p-kaczynski: Yeah, now that I think about it, it would probably be better to check what is used to query (id, uuid, source) and create partial indexes.

p-kaczynski: I did it and reverted from it for some reason, brain f...og.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/ServerConceptSchemeResource.java
p-kaczynski: Which, TBH, we don't.

garyttierney: We're not even returning 404's right now, so let's leave it out of this PR.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/infrastructure/event/ConceptEventListener.java
p-kaczynski: .NET has leaked in ;)

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
```diff
@@ -11,6 +12,17 @@
     private IdResolver() {
     }
 
+    public static Optional<UUID> resolve(String uri, Pattern pattern) {
+        if (uri == null)
```
garyttierney: Braces.

garyttierney: Also, `uri` should never be null as far as I can tell.

p-kaczynski:
Manual tests seems to show it works fine now. Running tests now, but due to expediency submitting for re-review

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
```diff
@@ -0,0 +1,53 @@
...
+
+        try {
+            var parsedUri = new URI(uri);
+            return resolve(parsedUri, pattern);
+        } catch (URISyntaxException ignored) {
+            return Optional.empty();
+        }
+    }
+    public static Optional<UUID> resolve(URI uri, Pattern pattern) {
+        if (uri == null || pattern == null) {
```
garyttierney: Why are we calling this if `uri` or `pattern` are null?

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/JdbcTemplateEx.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.sql.DataSource;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * An extension of JdbcTemplate that provides a queryForOptional, as a variant
+ * of queryForObject that doesn't throw if the queried object is not found.
+ */
+public class JdbcTemplateEx extends JdbcTemplate {
```
garyttierney: We don't need this implemented for `Optional` with every overload. A better solution is to just create a single static helper that can go from `List<T>` to `Optional<T>`.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/JdbcTemplateEx.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param results the result Collection (can be {@code null}
+     *                and is also expected to contain {@code null} elements)
+     * @return the single result object
+     * @throws IncorrectResultSizeDataAccessException if more than one
+     *                                                element has been found in the given Collection
+     */
+    private static <T> Optional<T> singleOptionalResult(@Nullable Collection<T> results) throws IncorrectResultSizeDataAccessException {
+        // This is identical to the nullableSingleResult implementation but returns
+        // Optional.empty() instead of throwing, if the results are empty
+        if (CollectionUtils.isEmpty(results)) {
```
garyttierney: We don't need `CollectionUtils`. A collection should never be `null`.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/management/ProjectModelRepository.java
```diff
@@ -49,9 +51,12 @@ public ProjectModel create(ProjectModel project) throws ProjectAlreadyExistsExce
      * @return the project with the given slug
      */
     @Transactional(Transactional.TxType.REQUIRED)
-    public ProjectModel find(String slug) {
-        ProjectDataSet dataSet = projectDao.loadDataSet(slug);
-        return projectMapper.map(dataSet);
+    public Optional<ProjectModel> find(String slug) {
+        var dataSet = projectDao.loadDataSet(slug);
```
garyttierney: Use a functional approach instead here:
```java
dataSet.map(dataSet -> projectMapper.map(dataSet))
```

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/JdbcTemplateEx.java
p-kaczynski: I don't trust them. Wake me up when non-nullables come ;)

Seriously though, I think the original template uses `@Nullable` for result collections.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
p-kaczynski: So all callers, current and future, can just call and get the answer.

- Can you try and resolve an id from this uri using this pattern? I just grabbed them and I have no idea what they are.
- Sure... nope, no id. You gave me lots of nulls.

Alternatively every caller of this method needs to do their checks, in the end leading to possible code duplication.

garyttierney: > Alternatively every caller of this method needs to do their checks, in the end leading to possible code duplication.

This is the correct approach. Allowing null everywhere, while having to test if you've been passed a `null`, is just nonsense.

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
p-kaczynski: It can be during the `create` calls.

Good call on braces. I might look into setting that as an error in InteliJ

taxonomy-manager-common/src/main/java/com/digirati/taxman/common/rdf/identity/IdResolver.java
garyttierney: If anything, it should throw an exception on `null` values. Not return a semantically correct result. Passing a null is a programmer error.

taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/storage/JdbcTemplateEx.java
p-kaczynski: That omits having to add another statement to each call, kind of keeps with the current flow&design. Also if ever JdbcTemplate gets such method, this can be very easily retired.

garyttierney: I agree that keeping the succinct code style in the various `XDao` classes is a good goal, but subclassing `JdbcTemplate` isn't the way to do it.

PR #47, opened 2019-11-20
DTM-148: Prevent Empty Title in Concept Schemes

p-kaczynski:


taxonomy-manager-rest-server/src/main/java/com/digirati/taxman/rest/server/taxonomy/validation/NonEmptyPlainLiteral.java
```diff
@@ -44,7 +45,9 @@ String message() default
     class Validator implements ConstraintValidator<NonEmptyPlainLiteral, Multimap<String, String>> {
         @Override
         public boolean isValid(Multimap<String, String> value, ConstraintValidatorContext context) {
-            return !value.isEmpty();
+            return !value.isEmpty()
+                    && value.keys().stream().allMatch(s -> s != null && s.length() > 0)
+                    && value.values().stream().allMatch(s -> s != null && s.length() > 0);
```
garyttierney: I think you only need to care about `value.values()` here.

p-kaczynski: I care about keys very much

PR #48, opened 2019-11-28
Add missing DROP FUNCTION/PROCEDURE directives

garyttierney:


garyttierney:
Some of these need both a DROP PROCEDURE and DROP FUNCTION because it used to be a function, but is now a procedure. Or vice versa.

PR #49, opened 2020-03-24
Use jgroups messaging for taxonomy updates

garyttierney:


taxonomy-manager-rest-server/src/main/resources/tcp.xml
```diff
@@ -0,0 +1,48 @@
...
+
+			enable_diagnostics="true"
+			diag_enable_udp="false"
+			diag_enable_tcp="true"
+
+			thread_pool.min_threads="0"
+			thread_pool.max_threads="20"
+			thread_pool.keep_alive_time="30000"/>
+
+	<TCPPING async_discovery="true"
```
danielgrant, line 27: I might be missing something, but isn't `TCPPING` a static configuration, whereby you need to know in advance the static IP addresses of the members of the cluster in advance? In this configuration, it looks like we're clustering with the local instance only and nothing else?

garyttierney: The `${initial_hosts:127.0.0.1[7800],127.0.0.1[7801]}"` bit gets templated with the value of `jgroups.initial_hosts` at runtime, which I was going to create a headless service for (like ELasticsearch).

danielgrant: Yeah I see that now. My bad.

Should there be a corresponding PR in `taxonomy-manager-infra` for the new service and environment variable passing to the deployment?

garyttierney: Yeah, though it depended on this approach being +1'd first. I'll open it now.

PR #54, opened 2020-06-15
Return the all token positions per concept from /analysis endpoint 

p-kaczynski:


taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/WordToken.java
```diff
@@ -39,6 +43,14 @@ public static boolean sharesCandidates(List<WordToken> a, List<WordToken> b) {
         return lexemes.values();
     }
 
+    public int getBeginPosition() {
```
garyttierney, line 46: New properties added, but equals/hsahCode not updated.

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/WordTokenSearchStrategy.java
```diff
@@ -13,4 +13,6 @@
     void unindex(WordTokenSearchEntry<IdT> entry);
 
     Set<IdT> match(List<WordToken> input);
+
+    Set<IdWithPosition<IdT>> matchWithPosition(List<WordToken> input);
```
garyttierney, line 17: Let's get rid of the default `match(...)` and rename `matchWithPosition(...)` to `match(...)`, making it the default.

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/nlp/corenlp/CoreNlpWordTokenizer.java
```diff
@@ -49,9 +53,13 @@ public static WordTokenizer create(String language) {
 
     @Override
     public List<WordToken> tokenize(String input) {
+        // Strip HTML tags and replace them with equivalent amount of '_' chars
+        HtmlTagReplacer.matcher(input).replaceAll(
```
garyttierney, line 57: We should enable the clean XML annotator instead of doing this. https://stanfordnlp.github.io/CoreNLP/cleanxml.html

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/nlp/corenlp/CoreNlpWordTokenizer.java
```diff
@@ -49,9 +53,13 @@ public static WordTokenizer create(String language) {
...
+                matchResult -> StringUtils.repeat('_', matchResult.end() - matchResult.start()));
+
         // Some input strings like to delimit text with a forward slash. This causes problems for Stanford CoreNLP,
         // which expects words to be delimited by whitespace or regular punctuation.
-        final String normalizedInput = input.replaceAll("/", " / ");
+        final String normalizedInput = input.replaceAll("/", " ");
```
garyttierney, line 62: This looks like it can potentially change matching semantics. Will leave that for regression testing to identify.

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/IdWithPosition.java
```diff
@@ -0,0 +1,25 @@
+package com.digirati.taxman.analysis;
+
+public class IdWithPosition<IdT> {
```
garyttierney, line 3: Let's give this a better class name. Something that identifies it as a "match". We should also store a reference to the matched tokens instead of just the begin/end positions.

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/nlp/corenlp/CoreNlpWordTokenizer.java
p-kaczynski: Will it work with malformed XML (like HTML)? Does it preserve the character position?

garyttierney: It should do. AFAIK it only searches for tags and doesn't try to parse it as a document.

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/nlp/corenlp/CoreNlpWordTokenizer.java
p-kaczynski: It was specifically done because the process looks like this:

CRUPlatform has an `Article (Title, Subtitle, Teaser, Body, CruViewComment)`

CRUPlatform creates a string like `Title + " " + Subtitle + " " + ... + CruViewComment`

CRUPlatform notes down the offset at which each "segment" starts in the constructed string

CRUPlatform sends string to TaxMan

TaxMan provides position(s) in the submitted string where the concept was found

CRUPlatform uses the previously saved offsets to translate the TaxMan position into "which field at which position" information

So the TaxMan _cannot_ modify the length of the text before it is tokenized.

p-kaczynski:
The `DCTerms.extent` was chosen after some consideration by me, but priority was getting that information _anywhere_ for the demo purposes. I am not in any way hung up on it. To be honest, none of the terms are purely appropriate, because the position is a result of the search, and they don't describe the concept itself, just its occurence. So in theory there should be a new transient resource `ConceptOccurence` that has the `Concept` as an attribute or relation, and holds the position.

However that would take longer, be more complex and ultimately not give us any benefits other than feeling of useless accomplishment ;)

taxonomy-manager-engine/src/main/java/com/digirati/taxman/analysis/nlp/corenlp/CoreNlpWordTokenizer.java
garyttierney: Hm, this is something we should look at supporting in another way in future. Happy with this for now.

garyttierney:
> So in theory there should be a new transient resource ConceptOccurence that has the Concept as an attribute or relation, and holds the position.

W3C Annotations ( ͡° ͜ʖ ͡°)

garyttierney (approved):

PR #58, opened 2020-11-02
Database project IDs, updated migrations

garyttierney:
Adds a creation listener that tacks on the project ID if present in a request to any `RdfModel`s that implement `ProjectScopedPersistentModel`. Still needs database changes.


p-kaczynski (approved):
LGTM, I see updated tests as well so I'm assuming it works at least in test.

